PR #1, opened 2020-01-31
Cache pathway

dehall:
(copied over from internal GitLab on behalf of @blangley28 )

Cache the pathway evaluation result. Previously the pathway was evaluated against a patient every time it loaded. This meant after viewing it on the explore screen it would re-evaluated when it was selected. Now the result is cached so it does not re-evaluate unless it has to.


danlee1025:
This looks good to me.  Pathways are no longer being re-evaluated when going from explore pathways screen or when switching between pathways.

danlee1025 (approved):

dehall:
GitHub is preventing me from clicking the approve button on this PR since I submitted it, but consider this comment as approval.

PR #2, opened 2020-01-31
Get fhir resource

dehall:
(copied from internal GitLab on behalf of @kghoreshi )

Adds info from fhir resources to the nodes. For now this is just pulling in the date from observations/procedures. You can check the documentation on nodes that aren't guidance nodes.

You can test this by clicking on a node and verifying that the date displayed is being pulled from the fhir resource from the fhir server.


src/components/Node/Node.tsx
```diff
@@ -54,9 +54,14 @@ const Node: FC<NodeProps> = ({
         <NodeIcon pathwayState={pathwayState} />
         {label}
       </div>
-      {isGuidanceState(pathwayState) && expanded && (
+      {expanded && (isGuidanceState(pathwayState) || documentation) && (
```
blangley28, line 57: Since this now renders for any node which has documentation it probably shouldn't be called RecNode anymore (since it is not just for recommendation nodes)

src/components/RecNode/RecNode.tsx
```diff
@@ -77,4 +66,71 @@ const RecNodeField: FC<RecNodeFieldProps> = ({ title, description }) => {
...
+            <FontAwesomeIcon icon="external-link-alt" className={classes.externalLink} />
+          </a>
+        </>
+      }
+    />,
+    <RecNodeField key="Code" title="Code" description={coding[0].code} />,
+    <RecNodeField key="Display" title="Display" description={coding[0].display} />
+  ];
+}
+function parseResource(resource: fhir.DomainResource): ReactElement[] {
```
blangley28: This looks easy enough to append to when needed but is there anything else we know should be included right now? The observation should probably include the system/code/display at the least

blangley28:
Looks good to me. Only last issue is that 12 unit tests fail

src/components/RecNode/RecNode.tsx
```diff
@@ -77,4 +66,71 @@ const RecNodeField: FC<RecNodeFieldProps> = ({ title, description }) => {
...
+  const returnValue: ReactElement[] = [];
+  switch (resource.resourceType) {
+    case 'Procedure': {
+      const procedure = resource as fhir.Procedure;
+      const start = procedure.performedPeriod && procedure.performedPeriod.start;
```
dehall: The Procedure profile also allows for procedure.performedDateTime

src/components/RecNode/RecNode.tsx
```diff
@@ -77,4 +66,71 @@ const RecNodeField: FC<RecNodeFieldProps> = ({ title, description }) => {
...
+      const end = procedure.performedPeriod && procedure.performedPeriod.end;
+      start &&
```
dehall: These should probably be traditional `if` statements rather than `&&` since the second part is a void function

src/components/Graph.tsx
```diff
@@ -27,7 +27,14 @@ const Graph: FC<GraphProps> = ({
   expandCurrentNode = true
 }) => {
   const graphElement = useRef<HTMLDivElement>(null);
-  const [path, setPath] = useState<string[]>([]);
+  const [pathwayResults, setPathwayResults] = useState<PathwayResults>({
+    path: [],
```
dehall: Is this default object necessary? It feels a little unusual to have an object populated with all empty fields rather than just use a null/undefined object

src/components/RecNode/RecNode.tsx
dehall: I can think of two things but not sure how we'll be modeling one of them. First one is MedicationRequests - we'd want the prescription date and status of medications. Second thing is the note that gets added, but we don't know the structure of that just yet. Beyond that, like you said anything else can be added pretty easily.

src/components/Node/Node.tsx
KeeyanGhoreshi: Will rename to DocNode

src/components/RecNode/RecNode.tsx
KeeyanGhoreshi: I'm going to leave additional fhir info for another task

KeeyanGhoreshi:
Should be good to go

src/components/Graph/Graph.tsx
```diff
@@ -137,15 +138,24 @@ const Graph: FC<GraphProps> = ({
             width: 400
           };
         } else {
+          // TODO: This obviously has to be changed eventually.
+          // The nodes height should change dynamically
+          const found =
+            evaluatedPathway &&
+            evaluatedPathway.pathwayResults &&
+            evaluatedPathway.pathwayResults.documentation.find(doc => {
+              return typeof doc !== 'string' && doc.state === e;
```
dehall, line 141: As discussed offline -- is there anything we can do here to ensure the node height is always at least large enough to accommodate the contents? Doesn't have to be perfect at this point so if it overshoots that's fine but better to have too much space than overlapping nodes.

src/components/RecNode/RecNode.tsx
dehall: Can you handle MedicationRequests here since that's going to be pretty common and you're already working this file? If we get that we probably won't need much more.

src/components/DocNode/DocNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+interface DocNodeProps {
+  pathwayState: GuidanceState | State;
+  isActionable: boolean;
+  documentation: DocumentationResource | undefined;
+  isGuidance: boolean;
+}
+
+const DocNode: FC<DocNodeProps> = ({ pathwayState, isActionable, documentation, isGuidance }) => {
+  let fhirFields: ReactElement[] = [];
+  if (documentation && documentation.resource) {
```
radamson:
```suggestion
  if (documentation?.resource) {
```

src/components/DocNode/DocNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+const DocNodeField: FC<DocNodeFieldProps> = ({ title, description }) => {
+  return (
+    <tr>
+      <td className={classes.descTitle}>{title}</td>
+      <td className={classes.desc}>{description}</td>
+    </tr>
+  );
+};
+
+function renderRecGuidance(pathwayState: GuidanceState): ReactElement[] {
```
radamson: should be Doc not Rec

radamson: I would make this a React FC (like DocNodeField).  That way we can still call this with JSX syntax.

See [React Conditional Rendering](https://reactjs.org/docs/conditional-rendering.html) or [functions returning JSX should be refactored into React components](https://github.com/yannickcr/eslint-plugin-react/issues/578)

i.e. `renderRecGuidance` vs `<RecGuidance/>`

src/components/DocNode/DocNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+}
+function parseResource(resource: fhir.DomainResource): ReactElement[] {
+  const returnValue: ReactElement[] = [];
+  switch (resource.resourceType) {
+    case 'Procedure': {
+      const procedure = resource as fhir.Procedure;
+      const start =
+        (procedure.performedPeriod && procedure.performedPeriod.start) ||
+        procedure.performedDateTime;
+      const end = procedure.performedPeriod && procedure.performedPeriod.end;
```
radamson:
```suggestion
      const end = procedure?.performedPeriod?.end;
```

src/components/DocNode/DocNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+    <DocNodeField key="Display" title="Display" description={coding[0].display} />
+  ];
+}
+function parseResource(resource: fhir.DomainResource): ReactElement[] {
+  const returnValue: ReactElement[] = [];
+  switch (resource.resourceType) {
+    case 'Procedure': {
+      const procedure = resource as fhir.Procedure;
+      const start =
+        (procedure.performedPeriod && procedure.performedPeriod.start) ||
```
radamson:
```suggestion
        (procedure?.performedPeriod?.start) ||
```

src/components/DocNode/DocNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+            <FontAwesomeIcon icon="external-link-alt" className={classes.externalLink} />
+          </a>
+        </>
+      }
+    />,
+    <DocNodeField key="Code" title="Code" description={coding[0].code} />,
+    <DocNodeField key="Display" title="Display" description={coding[0].display} />
+  ];
+}
+function parseResource(resource: fhir.DomainResource): ReactElement[] {
```
radamson, line 117: DocNode is starting to become responsible for a lot (with this function it has to know how to **parse and extract information from resources** AND **display them**).  Seems like we might want to think about breaking this up.

I think we need to think about this more to get the right solution, so I won't hold everything up on this, but it would be great from a reuse and maintainability perspective if we could break this up and out.

radamson (changes requested):
Reviewing a renamed file is harder than it needs to be.

dehall:
Looks good to me -- approved pending rebase

radamson (approved):

PR #3, opened 2020-01-31
default text

dehall:
(copied from internal GitLab on behalf of @kghoreshi )

Pretty simple.  If we want to do more with it I could make the default text pull in from a parent class or from the pathway.

### DIFF
--- a/src/components/RecNode/RecNode.tsx
+++ b/src/components/RecNode/RecNode.tsx
@@ -13,6 +13,10 @@ interface RecNodeProps {
 const RecNode: FC<RecNodeProps> = ({ pathwayState, isActionable }) => {
   const [comments, setComments] = useState<string>('');
   const resource = pathwayState.action[0].resource;
+  const defaultText =
+    `The patient and I discussed the treatment plan, ` +
+    `risks, benefits and alternatives.  The patient ` +
+    `expressed understanding and wants to proceed.`;
   const coding =
     'medicationCodeableConcept' in resource
       ? resource.medicationCodeableConcept.coding
@@ -46,11 +50,7 @@ const RecNode: FC<RecNodeProps> = ({ pathwayState, isActionable }) => {
             className={indexClasses.button}
             onClick={(e): void => {
               e.preventDefault();
-              setComments(
-                comments +
-                  // eslint-disable-next-line max-len
-                  `\nThe patient and I discussed the treatment plan, risks, benefits and alternatives.  The patient expressed understanding and wants to proceed.`
-              );
+              if (!comments.includes(defaultText)) setComments(comments + defaultText);
             }}
           >
             Use Default Text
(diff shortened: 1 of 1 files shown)


src/components/RecNode/RecNode.tsx
```diff
@@ -41,10 +42,24 @@ const RecNode: FC<RecNodeProps> = ({ pathwayState, isActionable }) => {
...
           <label>Comments:</label>
-          <button className={indexClasses.button} onClick={(e): void => e.preventDefault()}>
+          <button
+            className={indexClasses.button}
+            onClick={(e): void => {
+              e.preventDefault();
+              setComments(
+                comments +
+                  // eslint-disable-next-line max-len
+                  `\nThe patient and I discussed the treatment plan, risks, benefits and alternatives.  The patient expressed understanding and wants to proceed.`
```
radamson: Why the `\n`?

Screen Shot 2020-02-05 at 3 39 15 PM

src/components/RecNode/RecNode.tsx
```diff
@@ -41,10 +42,24 @@ const RecNode: FC<RecNodeProps> = ({ pathwayState, isActionable }) => {
       {isActionable && (
         <form className={classes.commentsForm}>
           <label>Comments:</label>
-          <button className={indexClasses.button} onClick={(e): void => e.preventDefault()}>
+          <button
+            className={indexClasses.button}
+            onClick={(e): void => {
+              e.preventDefault();
+              setComments(
+                comments +
```
radamson: I don't think you want to append the default text, but just use it.  If I click it multiple times it keeps adding which seems unwanted.

Screen Shot 2020-02-05 at 3 39 52 PM

src/components/RecNode/RecNode.tsx
```diff
@@ -41,10 +42,24 @@ const RecNode: FC<RecNodeProps> = ({ pathwayState, isActionable }) => {
...
+                  // eslint-disable-next-line max-len
```
radamson: Instead of disabling this eslint rule can you break this string over multiple lines?  Although I hate the look I think it's better we try to avoid selectively turning off the linter.

Use `+` or `\` (see [`Long literal strings` section in MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String))

radamson (changes requested):

src/components/RecNode/RecNode.tsx
KeeyanGhoreshi: We want to the text to append to already present text, not sure how to avoid the issue of having it fill the box multiple times on multiple clicks.

src/components/RecNode/RecNode.tsx
KeeyanGhoreshi: wanted to keep it separated from text already in the box, but I guess when it's alone it doesn't need it.

src/components/RecNode/RecNode.tsx
dehall: Yeah we do want it to append not totally replace. That's how the current design works. Quick & dirty option to not multiple append is `if (!comments.includes(defaultText)) setComments(comments + defaultText)`

KeeyanGhoreshi: sounds good to me

dehall:
Looks good -- this comment serves as my approval pending rebase

radamson (approved):
Looks good!

PR #4, opened 2020-01-31
Static demo endpoint

dehall:
(copied from internal GitLab on behalf of @danlee1025  )

*  Added `/demo` endpoint.
*  Added Maureen21 McOde090 as demo patient.
*  Added pathways to `public/static/pathways`.


src/components/App.tsx
```diff
@@ -62,35 +86,41 @@ const App: FC<AppProps> = ({ client }) => {
...
+          ></PathwaysList>
+        ) : (
+          <PatientView patientPathway={pathway} />
+        )}
+      </PathwayProvider>
+    );
+  };
+
+  return !demo ? (
+    <FHIRClientProvider client={client as PathwaysClient}>
```
dehall, line 151: I suspect the demo path will need some kind of FHIRClientProvider -- we're going to be referencing it downstream in the recommendation nodes via the accept/decline buttons, so if the client isn't present it might just crash. We don't want everything throughout the UI to be aware of whether it's in demo mode or not, but we also don't want to implement a whole static FHIR client that can stand in for the real thing.
Thoughts?

src/components/App.tsx
```diff
@@ -13,31 +13,55 @@ import PathwaysList from './PathwaysList';
 import { PathwayProvider } from './PathwayProvider';
 import { Pathway } from 'pathways-model';
 import useGetPathwaysService from './PathwaysService/PathwaysService';
+import FHIR from 'fhirclient';
+import demoRecords from '../fixtures/MaureenMcodeDemoPatientRecords.json';
```
dehall: Could these two demo files be combined into one, and then the Patient resource be parsed out only if necessary? One hope I had with this was that the patient could very easily be swapped out

src/components/App.tsx
dehall: For posterity - based on the discussion we just had offline, don't implement any mock FHIR client at this point since there's nothing that will actually exercise it. We can tackle that alongside the implementation of the real actions via the accept/decline button.

src/components/App.tsx
danlee1025: Removed demo patient and just implemented a search for the patient resource from demo records.

dehall:
Looks good to me. Again I can't approve since this I submitted it but consider this approval. Update: using Dan's approve button in place of my own. Next reviewer if you approve please merge.

danlee1025 (approved):

src/components/App.tsx
```diff
@@ -13,32 +13,56 @@ import PathwaysList from './PathwaysList';
 import { PathwayProvider } from './PathwayProvider';
 import { EvaluatedPathway } from 'pathways-model';
 import useGetPathwaysService from './PathwaysService/PathwaysService';
+import FHIR from 'fhirclient';
+import demoRecords from '../fixtures/MaureenMcodeDemoPatientRecords.json';
```
radamson:
```suggestion
import demoRecords from 'fixtures/MaureenMcodeDemoPatientRecords.json';
```

Use [absolute import](https://github.com/mcode/pathways/wiki/Code-Style-Guide#module-imports) unless the file is in the current or a child directory.

src/components/App.tsx
```diff
@@ -93,37 +117,48 @@ const App: FC<AppProps> = ({ client }) => {
...
-              callback={setEvaluatedPathwayCallback}
-              service={service}
-              resources={patientRecords}
-            ></PathwaysList>
-          ) : (
-            <PatientView evaluatedPathway={currentPathway} />
-          )}
-        </PathwayProvider>
-      </PatientProvider>
+  const renderPathwayProvider = (): ReactNode => {
```
radamson: I think this would be better as a FunctionalComponent (which will return a ReactNode) than a function returning a ReactNode.  That way we can use renderPathwayProvider as JSX in the return instead of calling a function.

[See NodeIcon in Node.tsx for an example of what I mean.](https://github.com/mcode/pathways/blob/master/src/components/Node/Node.tsx)

radamson: Also is there a better name for this than `renderPathwayProvider`?

src/components/PatientProvider.tsx
```diff
@@ -3,16 +3,19 @@ import { useFHIRClient } from './FHIRClient';
 
 interface PatientProviderProps {
   children: ReactNode;
+  demoPatient?: fhir.Patient;
```
radamson: should this just be `patient` I could imagine we might want to pass in the actual patient when initializing even when it is not a demo.

src/components/PatientProvider.tsx
```diff
@@ -3,16 +3,19 @@ import { useFHIRClient } from './FHIRClient';
...
 }
 
 export const PatientContext = createContext<fhir.Patient | null>(null);
```
radamson, line 9: Seems like the initial patient (`demoPatient`) should be set here when creating the context.

```suggestion
export const PatientContext = createContext<fhir.Patient | null>(demoPatient);
```

src/components/PatientProvider.tsx
```diff
@@ -3,16 +3,19 @@ import { useFHIRClient } from './FHIRClient';
...
 
-export const PatientProvider: FC<PatientProviderProps> = ({ children }) => {
+export const PatientProvider: FC<PatientProviderProps> = ({ children, demoPatient }) => {
   const client = useFHIRClient();
-  const [patient, setPatient] = useState<fhir.Patient | null>(null);
+  const [patient, setPatient] = useState<fhir.Patient | null>(demoPatient || null);
 
   useEffect(() => {
-    client.patient.read().then((patient: fhir.Patient) => setPatient(patient));
+    if (client && client.patient.read) {
```
radamson: Use optional chaining
```suggestion
    if (client?.patient?.read) {
```
You might be able to get away without the if statement too using this directly with `client?.patient?.read().then((patient: fhir.Patient) => setPatient(patient));`

radamson (changes requested):
Looks good.  A few minor comments.

src/components/App.tsx
danlee1025: This one is a little weird because I'm not actually creating a new component.  I just wanted to render the PathwayProvider component with the same exact children and props or else there would be a lot of duplication on lines 154 and 160.  Do you think I should still make a FunctionalComponent?

src/components/PatientProvider.tsx
danlee1025: When creating the context, it won't have access to the `demoPatient` prop since it is outside of the component.

src/components/App.tsx
danlee1025: Made this change.

src/components/App.tsx
radamson: I think its still worth making it a FunctionalComponent for the JSX syntax alone.  That way we can consistently use the JSX syntax and it is immediately clear to anyone that looks at it that a React Element is returned.  I rarely see the JSX that returns components with a function call like we do here (e.g. `foo()` vs `<Foo/>`).

Another way to think about it could be as a container vs presentational component.  This is still a component, but just a presentational subcomponent as we break everything down.

src/components/PatientProvider.tsx
radamson: good point

src/components/App.tsx
danlee1025: That makes sense.  Should be addressed now.

radamson (approved):

PR #8, opened 2020-02-06
Travis build deploy

dehall:
Adds a deploy step to the travis run script, based on instructions here: https://docs.travis-ci.com/user/deployment/pages/

Also adds the gh-pages URL to the package.json so that URLs in the build work as expected.

I tested this on a fork under my own account and it worked, see: https://github.com/dehall/pathways/tree/gh-pages and https://dehall.github.io/pathways/

The change to package.json and yarn.lock is to use the version of fhir-visualizers which is now on NPM, to fix this error: https://travis-ci.org/mcode/pathways/builds/646984479

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,14 @@ script:
   - yarn test
   - yarn build
   # replace the pathways service in the config file with the gh static location
+  # and the cql-to-elm service with a public one
   # -i means in-place
   # see https://stackoverflow.com/a/11245501
-  # note also multiline string here to get around some yaml weirdness
+  # note multiline string here gets converted into one line, not 3
   - >
       sed -i "/pathwaysService:/c\  pathwaysService: 'https://mcode.github.io/pathways/static/pathways/'," build/config.js
+      &&
+      sed -i "/cqlToElmWebserviceUrl:/c\  cqlToElmWebserviceUrl: 'https://cql-translation-service-zfk47ib3kq-uc.a.run.app/cql/translator'," build/config.js
 
 deploy:
   provider: pages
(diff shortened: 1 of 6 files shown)


radamson (approved):
LGTM.

We should stop tracking the `build` directory in fhir-visualizers now that it's on npm.

dehall:
Based on offline conversations with @blangley28 , I updated this so that it would work (as much as possible) via the GH pages. Unfortunately there's still a blocking error, and that's not something I can resolve in code or deployment scripts here.

Latest changes include:
 - changing the react-router to use a HashRouter instead of a BrowserRouter, so that the demo URL works on gh-pages. Because gh-pages uses a pretty simple URL matching scheme, trying to hit `/demo` would try to load an actual file /demo, and it wouldn't work. Now with a hash router, the demo URL is `/#/demo`. A little ugly but works.
 - Adding a static list of pathways that we can fetch using the existing pathway service. It's .html because of discrepancies between gh-pages and react-scripts. Added a note about that in a separate file
 - as part of deploy, rewrite the config file so that it points to the static list of pathways


So, the idea with all that is you should be able to hit https://dehall.github.io/pathways/#/demo or SMART launch to https://dehall.github.io/pathways/ and it will work. And it's so close. The last remaining piece is the CQL-to-ELM translation service. For one thing it's on our internal network, but it's also HTTP only, not HTTPS, so the browser is going to block requests on an HTTPS page from going to an HTTP resource. No way around that that I can see. I'm going to explore other possibilities around that, but for now there's nothing I can do about that and I don't think that should hold up this PR

blangley28 (approved):

PR #9, opened 2020-02-11
Branch node details

blangley28:
Clicking on branch nodes expands the node to provide more details. When the node has details from the CQL the value is populated into the node. If the data is missing a link appears to provide the missing information.
[PATHWAYS-146](https://jira.mitre.org/browse/PATHWAYS-146)


blangley28:
Screen Shot 2020-02-11 at 2 25 55 PM
Screen Shot 2020-02-11 at 2 26 08 PM

src/components/Node/Node.tsx
```diff
@@ -83,9 +82,13 @@ const NodeIcon: FC<NodeIconProps> = ({ pathwayState }) => {
   return <FontAwesomeIcon icon={icon} className={classes.icon} />;
 };
 
-function isGuidanceState(state: State): boolean {
+export function isGuidanceState(state: State): boolean {
   const { action } = state as GuidanceState;
   return action ? action.length > 0 : false;
 }
 
+export function isBranchState(state: State): boolean {
```
danlee1025: Since `Node` isn't the only component to use `isGuidanceState` and `isBranchState` we should maybe move these 2 functions into a utils file instead of exporting it from `Node`.

src/styles/_nodes.scss
```diff
@@ -4,9 +4,8 @@
   min-width: 400px;
   z-index: 1;
 
-  .docNode {
+  .ExpandedNode {
```
danlee1025: class names should be camelCase. .ExpandedNode -> .expandedNode

src/components/Node/Node.tsx
```diff
@@ -39,7 +39,7 @@ const Node: FC<NodeProps> = ({
   };
   const backgroundColorClass = isOnPatientPath ? classes.onPatientPath : classes.notOnPatientPath;
   const currentNodeClass = isCurrentNode ? classes.current : '';
-  const docNodeClass = isCurrentNode
+  const ExpandedNodeClass = isCurrentNode
```
danlee1025: this should be camelCase also.  `ExpandedNodeClass` -> `expandedNodeClass`

src/components/Node/Node.module.scss
```diff
@@ -38,13 +38,14 @@
   color: $black;
 }
 
-.docNode {
+.ExpandedNode {
```
danlee1025: This should be camelCase.

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -0,0 +1,195 @@
+import React, { FC, ReactNode, ReactElement } from 'react';
+import { GuidanceState, DocumentationResource, State } from 'pathways-model';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import classes from './ExpandedNode.module.scss';
```
danlee1025: According to style guide, CSS Module styles should be imported as `styles` so rename `classes` -> `styles`.

danlee1025 (changes requested):

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -0,0 +1,195 @@
...
+  if (documentation?.resource) {
+    switch (documentation.resourceType) {
+      case 'Observation': {
+        const observation = documentation.resource as fhir.Observation;
+
+        const valueCoding = observation.valueCodeableConcept?.coding;
+        valueCoding &&
+          returnElements.push(
+            <ExpandedNodeField
+              key="ValueSystem"
```
danlee1025, line 98: We probably want to use unique keys for these components.  If we have multiple branches expanded, these components would all have the same key value.  Maybe add the `pathwayState` label to the keys? I would change all the non-unique keys in `renderBranch` and `renderGuidance`.

radamson:
It seems like this is the second time expanded node has been renamed which makes it harder to review the changes that have been made to that file

`recNode` -> `docNode` -> `expandedNode`.

Is this the final file name?  it would be great if we could try and isolate these file name changes when we can, but I'm not sure the best way to do it...

src/styles/_nodes.scss
blangley28: Yup good catch. I used VsCode to rename but I guess it wasn't a case sensitive search and replace

src/components/ExpandedNode/ExpandedNode.tsx
blangley28: Can you point out where this is in the style guide? I can't find it and I noticed DropDown, Header, Navigation, Node, PathwayList, and PatientSnapshot use `classes` instead of `styles`. It seems like only Arrow uses `styles`

src/components/ExpandedNode/ExpandedNode.tsx
blangley28: I believe react keys only need to be unique within the component rendering them. In this case this is the ExpandedNode component which calls `renderBranch` and `renderGuidance`. Since multiple `ExpandedNodeField`s are rendered these all need a key. However I do not believe they need to be unique between nodes. To ensure they are unique within the `ExpandedNode` I prepended the keys for fields related to the value with "Value". This way instead of having two keys named "System" we instead have "System" and "ValueSystem". I believe this is enough, but correct me if I am misunderstanding

src/components/ExpandedNode/ExpandedNode.tsx
danlee1025: https://github.com/mcode/pathways/wiki/Code-Style-Guide#importing-styles

yeah I think we created the code style guide afterwards and never went back and changed the older files... but for new files we should probably follow the guide.

blangley28:
I agree it is difficult. It is also hard to say if it will change again. Originally the class was just for recommendation nodes, but it was then updated to show anything with documentation. Since this class already contained all the logic to show internal information on a node, it did not make sense to make a branch node class with the exact same logic. Instead I broadened the class once again to be expanded node. As far as I can tell, this is the most broad the class could become before it would just be called Node.

src/components/ExpandedNode/ExpandedNode.tsx
danlee1025: Ahh cool I think you're right and this should be fine.  I always thought they needed globally unique.

src/components/Node/Node.tsx
blangley28: Moved to a util module and exported for other modules to load

src/components/ExpandedNode/ExpandedNode.tsx
blangley28: Changed `classes` --> `styles` and `indexClasses` --> `indexStyles`
Updated in all files in PR 11 https://github.com/mcode/pathways/pull/11

danlee1025 (approved):

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  return (
+    <div className={indexStyles.expandedNode}>
+      <table className={styles.infoTable}>
+        <tbody>{guidance || branch}</tbody>
+      </table>
+      {isActionable && (
+        <form className={styles.commentsForm}>
+          <label>Comments:</label>
+          <button className={indexStyles.button} onClick={(e): void => e.preventDefault()}>
```
radamson: Looks like the default text functionality was removed.  This is an example of something difficult for git (or us) to catch when a file's name is changed, especially if there is a rebase involved.

https://github.com/mcode/pathways/blob/f6bb95a7622b01ed4cfb06a39e27b0c6b8e8320f/src/components/DocNode/DocNode.tsx#L41-L42

blangley28: Added back the default comments functionality. Also renamed ExpandedNode to DocNode for this PR

src/components/DocNode/__tests__/DocNode.test.tsx
```diff
@@ -56,7 +56,12 @@ const testMedicationRequestState: GuidanceState = {
 describe('<DocNode />', () => {
   it('renders a DocNode for action state', () => {
     const { getByText, queryByRole, queryByText } = render(
-      <DocNode pathwayState={testActionState} isActionable={false} isGuidance={true} />
+      <DocNode
+        pathwayState={testActionState}
+        isActionable={false}
+        isGuidance={true}
```
radamson, line 62: this prop was removed and is no longer necessary.

Ultimately I think the prop should stay though and this shouldn't change. (see comment in DocNode)

src/components/Node/Node.tsx
```diff
@@ -5,6 +5,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import classes from './Node.module.scss';
 import nodeClasses from 'styles/index.module.scss';
 import DocNode from 'components/DocNode';
+import { isGuidanceState } from 'utils/nodeUtils';
```
radamson, line 8: since `isGuidanceState` is still used in Node, we should just have it pass the result into DocNode and to know about this method.

if we really removed all use of `isGuidanceState` then I would agree with the prop removal from DocNode (and just having DocNode use it), but since we don't I think it's simpler passed in as a prop to DocNode.

src/components/DocNode/DocNode.tsx
```diff
@@ -2,41 +2,38 @@ import React, { FC, ReactNode, ReactElement, useState } from 'react';
...
 
 interface DocNodeProps {
   pathwayState: GuidanceState | State;
   isActionable: boolean;
   documentation: DocumentationResource | undefined;
-  isGuidance: boolean;
 }
 
-const DocNode: FC<DocNodeProps> = ({ pathwayState, isActionable, documentation, isGuidance }) => {
+const DocNode: FC<DocNodeProps> = ({ pathwayState, isActionable, documentation }) => {
```
radamson: I would add prop back and remove `isGuidanceState` import.  That way this component doesn't need to know about that method and we don't have to check in both DocNode and Node.

This is different from BranchState which Node doesn't care about (and therefore not a good candidate for lifting)

src/components/DocNode/DocNode.tsx
```diff
@@ -2,41 +2,38 @@ import React, { FC, ReactNode, ReactElement, useState } from 'react';
...
   return (
-    <div className={indexClasses.docNode}>
-      <table className={classes.infoTable}>
-        <tbody>
-          {isGuidance && <DocGuidance key="rec" pathwayState={pathwayState as GuidanceState} />}
-          {fhirFields}
-        </tbody>
+    <div className={indexStyles.docNode}>
+      <table className={styles.infoTable}>
+        <tbody>{guidance || branch}</tbody>
```
radamson: This would fit better with the principles of Thinking in React if we made `guidance` and `branch` actual react components (probably including the table).  Plus this file is getting pretty big and it'll be easier to read, test, and maintain.

then if would look something like `<NodeGuidance/>`.

This might be a decent amount of work though so I think it can wait till a future PR.

src/types/graph-model.d.ts
```diff
@@ -41,7 +41,7 @@ declare module 'graph-model' {
     points: Coordinate[];
   }
 
-  export interface ExpandedNodes {
+  export interface DocNodes {
```
radamson: This was already called ExpandedNodes in master, did we mean to rename it?  Just want to make sure we are not confusing `DocNode` (a react component) and `ExpandedNode` (an interface)

src/types/pathways-model.d.ts
```diff
@@ -5,7 +5,7 @@ declare module 'pathways-model' {
     library: string;
     criteria: Criteria[];
     states: {
-      [key: string]: GuidanceState | BranchState;
+      [key: string]: GuidanceState | BranchState | State;
```
radamson, line 8: since `GuidanceState` and `BranchState` both extend `State` this can just be `State`
```suggestion
      [key: string]: State;
```

src/types/graph-model.d.ts
radamson: If we do rename `DocNode` to `ExpandedNode` eventually I'd be concerned about continuing confusion between the two.

src/components/DocNode/DocNode.tsx
```diff
@@ -75,101 +72,133 @@ type DocNodeFieldProps = {
...
-        title="System"
+        key="value"
+        title="Value"
         description={
           <>
-            {coding[0].system}
-            <a href={coding[0].system} target="_blank" rel="noopener noreferrer">
-              <FontAwesomeIcon icon="external-link-alt" className={classes.externalLink} />
+            missing data
+            <a href={'example.com'} rel="noopener noreferrer">
```
radamson: why `example.com`?

src/components/DocNode/DocNode.tsx
```diff
@@ -75,101 +72,133 @@ type DocNodeFieldProps = {
...
+                    <FontAwesomeIcon icon="external-link-alt" className={styles.externalLink} />
+                  </a>
+                </>
+              }
+            />,
+            <DocNodeField key="ValueCode" title="Code" description={valueCoding[0].code} />,
+            <DocNodeField key="ValueDisplay" title="Display" description={valueCoding[0].display} />
+          );
+
+        const date = observation.effectiveDateTime;
```
radamson: assuming we are intentionally not handling the other choice types for effective[x]?  Fine as long as it is on our radar (could probably be its own PR anyways)

src/components/DocNode/DocNode.tsx
```diff
@@ -75,101 +72,133 @@ type DocNodeFieldProps = {
...
-      <DocNodeField key="Display" title="Display" description={coding[0].display} />
-    </>
-  );
-};
+    );
+  }
+  return returnElements;
+}
 
-// TODO: move this function somewhere more reasonable
```
radamson, line 116: Still think this should move at some point (i.e. figure out a new component to better represent this)

src/components/Graph/Graph.tsx
```diff
@@ -51,8 +57,8 @@ const Graph: FC<GraphProps> = ({
 
   // Get the layout of the graph
   const getGraphLayout = useCallback(
-    (expandedNodes: ExpandedNodes): Layout => {
-      return graphLayout(pathway, expandedNodes);
+    (docNodes: DocNodes): Layout => {
```
radamson: DocNodes here refers to the interface and we need to clear up that naming.

radamson (commented):
Tested it out and everything seemed to work.  I do have some concerns about the naming (DocNode vs ExpandedNode) to differentiate between the interface and React Components as right now they confusingly have the same name.

src/types/graph-model.d.ts
blangley28: You are correct. When I renamed back to DocNode I just used VSCode to replace all ExpandedNode with DocNode. Just to clarify the interface is called `ExpandedNodes` and the component `ExpandedNode`. So they are very similar (which is confusing), but still different

src/components/DocNode/DocNode.tsx
blangley28: It was complaining when it was empty but I don't have anything to put in there. Keeyan should be updating this in his branch where he actually implements this functionality

src/components/DocNode/DocNode.tsx
blangley28: This was moved from another part of the file. I think it came from Keeyan originally and he probably chose DateTime because that is the Synthea output.

src/components/DocNode/__tests__/DocNode.test.tsx
blangley28: Kept the prop so this does not need to change

src/components/Node/Node.tsx
blangley28: Node now computes isGuidance and passes the result into `NodeIcon` and `DocNode`

src/components/DocNode/DocNode.tsx
blangley28: Added back

src/types/graph-model.d.ts
blangley28: Renamed the interface to `NodeDimensions`

src/types/pathways-model.d.ts
blangley28: Undoing this commit. Breaks cql-extractor
Screen Shot 2020-02-17 at 3 35 42 PM

src/components/DocNode/DocNode.tsx
blangley28: Created [PATHWAYS-177](https://jira.mitre.org/browse/PATHWAYS-177)

radamson (approved):

src/components/DocNode/DocNode.tsx
blangley28: Created [PATHWAYS-178](https://jira.mitre.org/browse/PATHWAYS-178)

PR #10, opened 2020-02-12
Dynamic node height

danlee1025:
- Node height is no longer manually calculated.
- Passed ref to each Node so that Graph can get height of each rendered node.
- Modified layout algorithm to take `nodeDimensions` parameter.
- Layout is recalculated when window size changes and when a node toggles expansion.

image

### DIFF
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -1,8 +1,8 @@
 import React, { FC, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import graphLayout from 'visualization/layout';
-import { Node, NodeRef } from '../Node';
-import Arrow from '../Arrow';
+import Node from 'components/Node';
+import Arrow from 'components/Arrow';
 import { evaluatePatientOnPathway } from 'engine';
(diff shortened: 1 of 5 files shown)


danlee1025:
Added null check for resource in this PR to fix bug in PATHWAYS-176 since it was such a small change.

src/components/Graph/Graph.tsx
```diff
@@ -113,43 +123,15 @@ const Graph: FC<GraphProps> = ({
     }
   }, [expandCurrentNode, path, setExpanded]);
 
+  // Recalculate graph layout if window size changes or if a node is expanded
```
blangley28, line 126: Its not a huge deal because the graph layout algorithm is running really quickly but it looks like expanding a node causes the layout algorithm to run three times

blangley28 (approved):

src/components/Graph/Graph.tsx
danlee1025: True, this looks like it's also happening on `master` though.  I guess if the app gets really slow we can create a task to fix this.

dehall: We have a task for that: PATHWAYS-144

dehall (approved):

PR #11, opened 2020-02-14
Change classes to styles and kebab-case to camelCase

blangley28:
Updating all modules to use styles instead of classes and camelCase instead of kebab-case


danlee1025:
Thanks for doing this!  You forgot to update `classes` -> `styles` in `Dropdown` and `Header`.  There is still also some classes using kebab-case in `Header.module.scss`.

blangley28:
Good catch. Updated.

danlee1025 (approved):
Looks good!

dehall (approved):

PR #12, opened 2020-02-17
Confirmation and Missing Data Popups

KeeyanGhoreshi:
Adds both a popup for the accept buttons and for the "missing data ui" sections.  To test the missing data popup, use the patient "Miverfa" and open the tumor size node and click on missing data.

## This PR adds:
- An `ActionButton` component that can be used to create small or large accept and decline buttons with callbacks
- A `withConfirmationPopup` higher order component function that allows a confirmation popup to be added to components
- A `ConfirmedActionButton` which have the confirmation popup applied
- A `MissingDataPopup` for users to enter missing data

### Confirmation Popup
Screen Shot 2020-02-18 at 5 11 43 PM
### Missing Data Popup before selection
Screen Shot 2020-02-18 at 5 11 48 PM
### Missing Data Popup after selection
Screen Shot 2020-02-18 at 5 11 55 PM

## Test Metrics
branch |% Stmts | % Branch |  % Funcs |  % Lines |
------ |--------|----------|----------|----------|
master |  56.02 |    46.97 |    50.68 |    57.16 |
this PR|  57.28 |    48.25 |    52.99 |    58.45 |

For the relevant components:

File                                  |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
--------------------------------------|----------|----------|----------|----------|-------------------|
All files                             |    57.03 |    47.98 |    52.36 |    58.19 |                   |
 src/components/ActionButton          |      100 |      100 |      100 |      100 |                   |
  ActionButton.tsx                    |      100 |      100 |      100 |      100 |                   |
 src/components/ConfirmedActionButton |      100 |      100 |      100 |      100 |                   |
  ConfirmedActionButton.ts            |      100 |      100 |      100 |      100 |                   |
 src/components/MissingDataPopup      |    53.33 |       25 |       50 |    53.33 |                   |
  MissingDataPopup.tsx                |    53.33 |       25 |       50 |    53.33 |... 50,52,53,64,66 |
 src/components/PathwayPopup          |      100 |    83.33 |      100 |      100 |                   |
  PathwayPopup.tsx                    |      100 |    83.33 |      100 |      100 |                33 |
 src/components/withConfirmationPopup |      100 |      100 |      100 |      100 |                   |
  withConfirmationPopup.tsx           |      100 |      100 |      100 |      100 |                   |


src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+const ActionButton: FC<ActionButtonProps> = ({ type, size, onClick }) => {
+  switch (size) {
+    case 'small':
+      return <SmallActionButton type={type} onClick={onClick} />;
+    case 'large':
+      return <LargeActionButton type={type} onClick={onClick} />;
+  }
+};
+
+const ConfirmedActionButton: FC<ActionButtonProps> = withConfirmationPopup(
```
blangley28: Lint complains about this 
/Users/blangley/Documents/repos/pathways/src/components/ActionButton/ActionButton.tsx
  43:76  error  Replace `⏎··ActionButton⏎` with `ActionButton`  prettier/prettier

```suggestion
const ConfirmedActionButton: FC<ActionButtonProps> = withConfirmationPopup(ActionButton);
```

src/components/ActionButton/ActionButton.module.scss
```diff
@@ -0,0 +1,23 @@
...
+  font-weight: 800;
+  margin-right: 0.5em;
+}
+
+.check {
+  @extend .selectButton;
+  background: $green;
+}
+
+.times {
```
blangley28: I assume this class was labeled `times` because the icon is called that. But since we are using it as an X or close button I think it would make more sense to rename it based on how it is being used

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { FC } from 'react';
+import classes from '../ExpandedNode/ExpandedNode.module.scss';
+import indexClasses from '../../styles/index.module.scss';
+import withConfirmationPopup from '../withPopup/withConfirmationPopup';
```
blangley28: According to the style guide these should be called `expandedNodeStyles` and `indexStyles` and use an absolute import

src/components/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+import styles from './ActionButton.module.scss';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faTimes, faCheck } from '@fortawesome/free-solid-svg-icons';
```
blangley28: All of the font awesome icons are loaded in `utils/fontawesomeLibrary.js`

src/components/MissingDataPopup/MissingDataPopup.module.scss
```diff
@@ -0,0 +1,65 @@
...
+    text-align:center;
+    border-radius: 15%;
+    font-weight: 800;
+
+    &:hover{
+        background: #222;
+    }
+
+    &.selected{
+        background: rgb(94, 138, 160);
```
blangley28: Is this a color which may be used elsewhere? We might want to define it as a variable in styles

src/components/MissingDataPopup/MissingDataPopup.module.scss
```diff
@@ -0,0 +1,65 @@
...
+  text-align:center;
+  border-radius: 15%;
+  font-weight: 800;
+
+  &.check {
+    background: rgb(56, 167, 117);
+    margin-right: 5px;
+
+    &:hover{
+      background: rgb(77, 196, 142);
```
blangley28: I would defer to @noranda but I think for hover styling we are doing darken 10%

src/components/MissingDataPopup/MissingDataPopup.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { FC, useState } from 'react';
+import styles from './MissingDataPopup.module.scss';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import PathwayPopup from '../PathwayPopup/PathwayPopup';
+import nodeClasses from '../ExpandedNode/ExpandedNode.module.scss';
+import ActionButton from '../ActionButton/ActionButton';
```
blangley28: Same thing here - these should be absolute not relative imports and the scss module should be imported as `nodeStyles`

src/components/withPopup/withConfirmationPopup.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { FC } from 'react';
+import styles from './withConfirmationPopup.module.scss';
+import PathwayPopup from 'components/PathwayPopup';
+import ActionButton from '../ActionButton/ActionButton';
```
blangley28: Use absolute import

src/components/ActionButton/ActionButton.module.scss
radamson: renamed to `accept` and `decline` to better indicate how they are used.

src/components/MissingDataPopup/MissingDataPopup.module.scss
radamson: 👍 happy to tweak the value if desired.

src/components/ActionButton/ActionButton.tsx
radamson: I disagree with the approach because it allows the app to silently fail if they haven't been imported, which is also hard to know without using the explicit import.  Also it looks like the explicit inclusion might allow for a smaller bundle size.

Regardless, I changed it to use the library import until we can change everything over.

src/components/ActionButton/__tests__/ActionButton.test.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import ActionButton from '../ActionButton';
+
+const buttonText = 'Accept';
+const mockCallback = jest.fn();
+
+describe('<ActionButton />', () => {
```
blangley28, line 8: Does this only test the large action button because it has text and is therefore easier to get the element? Or was only one tested because they are so similar?

src/components/ConfirmedActionButton/ConfirmedActionButton.ts
```diff
@@ -0,0 +1,6 @@
+import withConfirmationPopup from '../withConfirmationPopup';
+import ActionButton from '../ActionButton/ActionButton';
```
blangley28: Again use absolute import if not in current directory

src/components/ActionButton/__tests__/ActionButton.test.tsx
radamson: because it had text and was easier to get.  I considered adding more tests for the different permutations, but decided to focus on adding more test elsewhere to cover other gaps.

src/components/ConfirmedActionButton/__tests__/ConfirmedActionButton.test.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import { ConfirmedActionButton } from '../ConfirmedActionButton';
```
blangley28, line 3: Absolute import again

radamson: we use relative import in tests (unless we are importing a component that is not the primary component being tested)

src/components/PathwayPopup/__tests__/PathwayPopup.test.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import PathwayPopup from '../PathwayPopup';
```
blangley28, line 3:
```suggestion
import PathwayPopup from 'components/PathwayPopup';
```

src/components/withConfirmationPopup/__tests__/withConfirmationPopup.test.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { FC } from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import withConfirmationPopup from 'components/withConfirmationPopup';
```
blangley28: This uses an absolute import

yarn.lock
```diff
@@ -3170,11 +3195,6 @@ boolbase@^1.0.0, boolbase@~1.0.0:
   resolved "https://registry.yarnpkg.com/boolbase/-/boolbase-1.0.0.tgz#68dff5fbe60c51eb37725ea9e3ed310dcc1e776e"
   integrity sha1-aN/1++YMUes3cl6p4+0xDcwed24=
 
-bootstrap@^4.4.1:
```
blangley28, line 3173: Is there a reason bootstrap was removed. I think it is a dependency for the fhir visualizers

src/components/ConfirmedActionButton/__tests__/ConfirmedActionButton.test.tsx
blangley28: Cool. Could you please update the code style guide to mention that

blangley28:
Looks good to me!

blangley28 (approved):

yarn.lock
radamson: only thing that changed in the `package.json` was the addition of semantic-ui.  can't speak for why that changed...

any ideas @dehall?  Maybe it is related to getting `fhir-visualizers` through npm now (I'd imagine it should bring whatever dependencies it needs without us having to specify)

dehall: No clue. fhir-visualizers has been a pain to get working correctly. I'll try a fresh clone of this branch and see if it loads correctly.

dehall: So no it's not using bootstrap as expected -- the patient record UI would be more consistently spaced and have row-level highlighting on hover if it did -- though it's also not failing to load the app completely like we've seen before when bootstrap wasn't included. I suspect it's an issue with the fhir-visualizers npm package, not with us using it here. Also we're going to get a new design for the patient record on seen anyway, so we can investigate further when we work on that new design.

dehall:
So it looks like the user can update missing data on states other than the current state, for example looking ahead in a pathway with the demo patient: 
image

On one hand, I'd say there's no reason to explicitly block this, but because this state isn't on the patient path, is it possible the patient's record does actually have the requisite data but it doesn't make it to the node? In this specific example, the patient has an ER negative observation so the branch above shouldn't really be missing data.
I'm going to keep digging but if anyone else has thoughts please chime in

blangley28:
I spoke to Zeshan and Jim about this - the consensus we came to was future states should either show the patient record if it already exists or allow the Oncologist to input missing data

dehall:
Ok, so it looks like we'll have to change the engine/data model to do that. The PathwayResults object that we get back has `path` and `documentation` as arrays of the same length so the node referenced at `path[i]` relates to to `documentation[i]`.  For a state not on the path, we don't get documentation for it. 
That's a bigger change and not purely visual, so that doesn't have to be addressed here. I'll put that in Jira.

dehall:
Upon closer inspection, that summary might not be completely right. But the key point is the engine doesn't return documentation for nodes further up in the pathway.

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -130,18 +128,15 @@ function renderBranch(documentation: DocumentationResource | undefined): ReactEl
       }
     }
   } else {
+    const values: string[] = pathwayState.transitions.map(transition => {
+      const description = transition.condition?.description;
+      return description ? description : '';
```
dehall, line 133: Not worth changing by itself but does our babel setup support the null coalescer? `??` , ex `description = transition.condition?.description ?? ''`

dehall (approved):

src/components/ExpandedNode/ExpandedNode.tsx
danlee1025: It should since this is used in `Graph.tsx`

PR #13, opened 2020-02-17
In-Progress Nodes

dehall:
Adds a status field and icon to nodes indicating the status. Accepted nodes should be marked with a check, declined nodes should be marked with an X, and nodes that have already had an action taken on them should not be highlighted in red.

Note that we currently don't have the ability to accept/decline, so the logic for now is to assume anything actionable on the patient path with documentation was accepted, hence I don't think it's actually possible for anything to appear as declined yet.

### Samples

With the hardcoded demo patient:
image
image

With "Magenta273 McOde090", who is based on Maureen but has a chemotherapy regimen started with status  "in progress"
image
image

### DIFF
--- a/src/components/ExpandedNode/ExpandedNode.tsx
+++ b/src/components/ExpandedNode/ExpandedNode.tsx
@@ -23,7 +23,7 @@ const ExpandedNode: FC<ExpandedNodeProps> = ({
 }) => {
   const [comments, setComments] = useState<string>('');
 
-  const status = documentation?.resource && renderStatusFields(documentation);
+  const status = documentation?.resource && renderStatusField(documentation);
   const guidance = isGuidance && renderGuidance(pathwayState as GuidanceState, documentation);
   const branch = isBranchState(pathwayState) && renderBranch(documentation, pathwayState);
 
@@ -83,16 +83,12 @@ const ExpandedNodeField: FC<ExpandedNodeFieldProps> = ({ title, description }) =
   );
 };
 
-function renderStatusFields(documentation: DocumentationResource): ReactElement[] {
-  const returnValue: ReactElement[] = [];
-
+function renderStatusField(documentation: DocumentationResource): ReactElement {
   const status = documentation.status;
 
   const rawDate = documentation.resource?.meta?.lastUpdated;
   const date = rawDate && new Date(rawDate).toLocaleString('en-us');
-  returnValue.push(<ExpandedNodeField key="Status" title={status} description={date} />);
-
-  return returnValue;
+  return <ExpandedNodeField key="Status" title={status} description={date} />;
 }
 
 function renderBranch(
--- a/src/components/ExpandedNode/ExpandedNode.tsx
+++ b/src/components/ExpandedNode/ExpandedNode.tsx
@@ -23,6 +23,7 @@ const ExpandedNode: FC<ExpandedNodeProps> = ({
 }) => {
   const [comments, setComments] = useState<string>('');
 
+  const status = documentation?.resource && renderStatusFields(documentation);
   const guidance = isGuidance && renderGuidance(pathwayState as GuidanceState, documentation);
   const branch = isBranchState(pathwayState) && renderBranch(documentation, pathwayState);
 
@@ -34,7 +35,10 @@ const ExpandedNode: FC<ExpandedNodeProps> = ({
   return (
     <div className={indexStyles.expandedNode}>
       <table className={styles.infoTable}>
-        <tbody>{guidance || branch}</tbody>
+        <tbody>
+          {status}
+          {guidance || branch}
+        </tbody>
       </table>
       {isActionable && (
         <form className={styles.commentsForm}>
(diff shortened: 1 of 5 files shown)


src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -83,6 +87,18 @@ const ExpandedNodeField: FC<ExpandedNodeFieldProps> = ({ title, description }) =
...
 };
 
+function renderStatusFields(documentation: DocumentationResource): ReactElement[] {
+  const returnValue: ReactElement[] = [];
+
+  const status = documentation.status;
+
+  const rawDate = documentation.resource?.meta?.lastUpdated;
+  const date = rawDate && new Date(rawDate).toLocaleString('en-us');
+  returnValue.push(<ExpandedNodeField key="Status" title={status} description={date} />);
```
blangley28: Are we expecting to have other ExpandedNodeFields being returned by this function?

blangley28:
It seems like this implies a node could be the current node but not actionable? Is this the case? I know it is right now since there is no write back on accepting a node but in reality if the node has been accepted there will be a new current node

dehall:
Yeah that's the idea -- like if there's a MedicationRequest for a regimen that hasn't completed yet, or a ServiceRequest for a procedure that hasn't been performed yet. The engine already supports it, and I had a test patient to demo it but I forget who that was. I can create a new one to show that.

src/components/ExpandedNode/ExpandedNode.tsx
dehall: Not sure. Originally I had it as two fields "status" and "status date" but then the design had it more like this. I can turn it back into a single item instead of an array if we confirm that's what we want for the design

dehall: Changed it to return just the single field not an array

src/components/ExpandedNode/__tests__/ExpandedNode.test.tsx
```diff
@@ -138,6 +138,11 @@ describe('<ExpandedNode />', () => {
       />
     );
 
+    expect(getByText(testDoc.status)).toBeVisible();
+    // date string from above. TODO: brittle
+    const date = new Date('2020-02-10T18:55:18.991+00:00').toLocaleString();
```
radamson: I would just make this a constant at the top right above the `testDoc` declaration and have both `testDoc` and this use it.

Not perfect, but a little less brittle and DRYer.

src/components/Node/Node.tsx
```diff
@@ -31,6 +33,10 @@ interface NodeIconProps {
   isGuidance: boolean;
 }
 
+interface StatusIconProps {
```
radamson: I know the `NodeIconProps` doesn't follow this rule, but it would make sense for this declaration to be moved down closer to where it is used (e.g. right above `StatusIcon`).  If we used it in multiple places I would agree with having it at the top, but it's really for the sub-component.

bonus points if you move `NodeIconProps` too.

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -79,6 +83,14 @@ const ExpandedNodeField: FC<ExpandedNodeFieldProps> = ({ title, description }) =
   );
 };
 
+function renderStatusField(documentation: DocumentationResource): ReactElement {
```
radamson: Make this a React Component.

We should consistently use JSX and think in React components. We have PATHWAYS-177 which came out of similar discussions in https://github.com/mcode/pathways/pull/9 regarding `renderBranch` and `renderGuidance`, but we should stop now.  Right now every time `ExpandedNode` is called or a re-render happens a new instance of this function will be created (bleh!).

This should be a React component similar to [`NodeIcon` in Node.tsx](https://github.com/mcode/pathways/blob/master/src/components/Node/Node.tsx#L87) or the `StatusIcon` component you also just added in this PR below 🙂  .

_Note that it will have to be defined outside of `ExpandedNode` so that it doesn't get redefined on every render._

src/components/Node/Node.tsx
```diff
@@ -50,30 +56,42 @@ const Node: FC<NodeProps & { ref: Ref<HTMLDivElement> }> = forwardRef<HTMLDivEle
...
+        ? styles.childOnPatientPath
+        : styles.childNotOnPatientPath;
+    }
     const isGuidance = isGuidanceState(pathwayState);
+    // TODO: how do we determine whether a node has been accepted or declined?
+    // for now:
+    // if it's a non-actionable guidance state on the path: accepted == has documentation
+    // if it's actionable, not guidance or not on the path: null
+    const isAccepted =
+      isOnPatientPath && isGuidance && !isActionable ? documentation != null : null;
```
radamson: it would be more readable to break this into two parts.

One for `const nodeState = isOnPatientPath && isGuidance && !isActionable` and another for `nodeState ? documentation !=null : null`

the `nodeState` name I put in is not great though.

src/styles/_nodes.scss
```diff
@@ -14,10 +14,26 @@
...
+    justify-content: flex-end;
   }
 }
 
 .nodeTitle {
   padding: 15px;
+  /* use the entire width of its parent 
+     so we can align something to the right */
+  width: 100%;
+  display: flex;
```
radamson, line 26: nice use of flexbox 👍

src/styles/_nodes.scss
```diff
@@ -14,10 +14,26 @@
...
 }
+
+.iconAndLabel {
+  flex-grow: 1;
+  text-align: center;
+}
+
+.statusIcon {
+  margin-left: auto;
+  padding-left: 10px;
```
radamson: should we use `em` instead of `px` here to make the padding respond to text size?

radamson (commented):
Nice work!

Left a few comments for you to consider.

dehall:
@radamson all comments should be addressed in the latest

radamson (approved):
Looks good 👍

danlee1025 (approved):
LGTM

PR #14, opened 2020-02-19
Use Explicit Import for Icons

radamson:
This PR implements the [explicit import of Font Awesome icons](https://github.com/FortAwesome/react-fontawesome#usage) (replacing the current library approach)

- It prevents accidentally using an icon that hasn't been loaded.  Otherwise the icon would just silently not appear.  you can test this by setting one the the existing icons to something like `'address-book'` which is a valid font awesome icon, but not currently loaded on `master`.
- Prevents unused icons from hanging around.  Otherwise you'd have to check to make sure it isn't used anywhere before deleting it from the library.
- Makes the bundle size a little smaller (1 kB, which isn't a big deal)
- To add an icon you only have to include it where you use it and not update or check multiple places.

I'll update the style guide if we accept and merge for guidance on using icons.


dehall (approved):

dehall:
This approach makes more sense to me than the library option

danlee1025 (approved):

PR #17, opened 2020-02-24
Add New Neoadjuvant HER2+ Pathway

blangley28:
Adds the latest HER2+ Early Stage pathway. Added to the static directory to be accessible via the demo endpoint.
Screen Shot 2020-02-24 at 12 31 28 PM
Screen Shot 2020-02-24 at 12 31 39 PM

### DIFF
--- a/public/static/cql/upenn_her2_library.cql
+++ b/public/static/cql/upenn_her2_library.cql
@@ -39,12 +39,16 @@ code "Abnormal estradiol (finding) code": '131064004' from "SNOMEDCT" display 'A
 
 // SNOMED Procedures
 code "Chemotherapy (procedure) code": '367336001' from "SNOMEDCT" display 'Chemotherapy (procedure)'
+code "Lumpectomy of breast (procedure) code": '392021009' from "SNOMEDCT" display 'Lumpectomy of breast (procedure)'
 code "Oophorectomy (procedure) code": '83152002' from "SNOMEDCT" display 'Oophorectomy (procedure)'
 
 
 // RXNORM Medication Request
+code "adotrastuzumab emtansine code": '1371041' from "RXNORM" display 'ado-trastuzumab emtansine'
 code "trastuzumab code": '224905' from "RXNORM" display 'trastuzumab'
 code "anastrozole code": '84857' from "RXNORM" display 'anastrozole'
+code "pertuzumab code": '1298944' from "RXNORM" display 'pertuzumab'
+code "paclitaxel code": '56946' from "RXNORM" display 'paclitaxel'
 code "tamoxifen code": '10324' from "RXNORM" display 'tamoxifen'
 code "Lupron code": '203852' from "RXNORM" display 'Lupron'
 
--- a/public/static/pathways/upenn_her2_early_stage_pathway.json
+++ b/public/static/pathways/upenn_her2_early_stage_pathway.json
@@ -1,6 +1,6 @@
 {
     "name": "HER2+ Early Stage Pathway",
-    "description": "Pathway for the UPenn Early Stage HER2+ Breast Cancer Pathway",
+    "description": "Pathway for the Early Stage HER2+ Breast Cancer Pathway",
     "library": "mCODE_Library.cql",
     "version": 1.0,
     "criteria": [
@@ -111,7 +111,7 @@
             }
           }
         ],
-        "cql": "[Procedure: \"Lumpectomy of breast (procedure) code\"] Lumpectomy return Tuple{ resourceType: 'Procedure', id: Lumpectomy.id.value, status: Lumpectomy.status.value }",
+        "cql": "if [Procedure: \"Lumpectomy of breast (procedure) code\"] then [Procedure: \"Lumpectomy of breast (procedure) code\"] Lumpectomy return Tuple{ resourceType: 'Procedure', id: Lumpectomy.id.value, status: Lumpectomy.status.value } else [ServiceRequest: \"Lumpectomy of breast (procedure) code\"] Request return Tuple{ resourceType: 'Service[Procedure: \"Teleradiotherapy procedure (procedure) code\"] Radiation return Tuple{ resourceType: 'Procedure', id: Radiation.id.value, status: Radiation.status.value } ",
         "transitions": [
           {
             "transition": "PaclitaxelTrastuzumab"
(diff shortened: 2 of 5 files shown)


public/static/pathways/upenn_her2_early_stage_pathway.json
```diff
@@ -0,0 +1,589 @@
...
+        ]
+      },
+      "Surgery1": {
+        "label": "Surgery",
+        "action": [
+          {
+            "type": "create",
+            "description": "Cancer Related Surgery Procedure",
+            "resource": {
+              "resourceType": "Procedure",
```
dehall: I'm not sure about using `Procedure` resources in the action here. (Understood that we have `Procedure`s in the older pathways)
On one hand I think we should be using only "Request" type resources in the `action` field, for example in this case a `ServiceRequest`(R4)/`ProcedureRequest`(STU3/DSTU2). When the oncologist clicks the accept button here, they aren't saying the surgery is happening right now or already done, they are requesting the action to happen.  
But on the other hand that could potentially also be represented as a `Procedure` with `status`: `preparation`.
Let's discuss after standup today.

Also yes I'm aware using `ServiceRequest` here makes the CQL more difficult as we'd then want to check for either the proecedure request or performed procedure. Maybe there's something we could extract as a helper function to make that easier.

blangley28: Hmm that is a good point. Could we not do a Request type for the resource and then keep the CQL to check for the actual procedure to be complete? I don't think there is a link between the CQL and the action resource type

dehall: We want the CQL to be able to recognize whether the node is not started at all, in-progress, or completed, so I think the CQL has to look for both, or if just one, then the request. If the user clicks on the accept button and the app writes back a request, we want to recognize the next time that that they clicked that button and the node is at least in-progress.

blangley28: From a quick meeting with Dylan and Reece this branch is going to modify the CQL of the action states to return either the service request or the actual procedure if it exists. From there a future task can mark nodes progress based on the status of the returned documentation resource.

public/static/pathways/upenn_her2_early_stage_pathway.json
```diff
@@ -0,0 +1,589 @@
...
+                    "display": "trastuzumab"
+                  }
+                ],
+                "text": "trastuzumab"
+              }
+            }
+          }
+        ],
+        "cql": "[MedicationRequest: \"pertuzumab code\"] MR return Tuple{ resourceType: 'MedicationRequest', id: MR.id.value , status: MR.status.value} ",
+        "transitions": []
```
radamson: does this have to be here if there are no transitions?  We should not be including empty arrays.  If it is a problem with an engine we should create a ticket to address that.

public/static/cql/upenn_her2_library.cql
```diff
@@ -39,12 +39,16 @@ code "Abnormal estradiol (finding) code": '131064004' from "SNOMEDCT" display 'A
...
 code "Chemotherapy (procedure) code": '367336001' from "SNOMEDCT" display 'Chemotherapy (procedure)'
+code "Lumpectomy of breast (procedure) code": '392021009' from "SNOMEDCT" display 'Lumpectomy of breast (procedure)'
 code "Oophorectomy (procedure) code": '83152002' from "SNOMEDCT" display 'Oophorectomy (procedure)'
 
 
 // RXNORM Medication Request
+code "adotrastuzumab emtansine code": '1371041' from "RXNORM" display 'ado-trastuzumab emtansine'
 code "trastuzumab code": '224905' from "RXNORM" display 'trastuzumab'
 code "anastrozole code": '84857' from "RXNORM" display 'anastrozole'
+code "pertuzumab code": '1298944' from "RXNORM" display 'pertuzumab'
```
radamson, line 50: How were these codes picked?  This TTY for this code is IN, and it would be better to have a code like
`1298946 (pertuzumab Injectable Product)` which has a TTY of SCDG to better meet the mCODE/Argonaut/US Core IGs.

public/static/pathways/upenn_her2_early_stage_pathway.json
```diff
@@ -0,0 +1,589 @@
+{
+    "name": "HER2+ Early Stage Pathway",
+    "description": "Pathway for the Early Stage HER2+ Breast Cancer Pathway",
+    "library": "mCODE_Library.cql",
+    "version": 1.0,
```
radamson: Is this just an arbitrary version number?  I didn't notice a version on any of the other pathways.

public/static/pathways/upenn_her2_early_stage_pathway.json
```diff
@@ -0,0 +1,589 @@
...
+    "criteria": [
+      {
+        "elementName": "Condition",
+        "expected": "Breast Cancer",
+        "cql": "\"Primary Cancer Condition\" PrimaryCancer return Tuple{ value: PrimaryCancer.code.text.value, match: PrimaryCancer.code.text.value ~ 'Malignant neoplasm of breast (disorder)' } "
```
radamson: Let me preface this by saying I'm a CQL newbie.

I noticed that the quoted part of `"\"Primary Cancer Condition\" PrimaryCancer ` is not surrounded by `[]`.  Should it be?

e.g. on line 41 I see `"[Observation: \"Primary tumor.clinical [Class] Cancer code\"] TLarge`.

radamson (commented):
Nice.  Looks good to me and I was able to pull it up in the app and check it out.  I left a couple questions below.

public/static/pathways/upenn_her2_early_stage_pathway.json
blangley28: Yes the data model requires transitions to be there. Created [PATHWAYS-193](https://jira.mitre.org/browse/PATHWAYS-193)

public/static/cql/upenn_her2_library.cql
blangley28: I'm not sure what any of those acronyms mean... but the codes were all selected from RxNav. I searched the drug name and took the code listed. Are you suggesting to select one of the specific items instead of just the generic drug code?

public/static/pathways/upenn_her2_early_stage_pathway.json
blangley28: Yes it is arbitrary. I can remove it. The reason it exists is because I started off with one of the modified HER2+ pathways which was on sharepoint not on the repo (because it wasn't being used yet). To distinguish between the original version and the new version I added the version tag.

blangley28: Yes it is arbitrary. I can remove it. The reason it exists is because I started off with one of the modified HER2+ pathways which was on sharepoint not on the repo (because it wasn't being used yet). To distinguish between the original version and the new version I added the version tag.

public/static/pathways/upenn_her2_early_stage_pathway.json
blangley28: Nope the [] are not needed. So the `Primary Cancer Condition` is defined within the library. This is the same CQL as the other her2+ pathway so you can verify it works there too.

public/static/cql/upenn_her2_library.cql
radamson: Talked with @blangley28 and @dehall about this and it sounds like we'll go with the SCD representations.  Need to follow up with Jim and Zeshan to make sure we get it right though.

Created PATHWAYS-194 to address this.

radamson (approved):

danlee1025 (approved):

PR #21, opened 2020-03-03
App interactivity

danlee1025:
Putting this up to get some initial feedback.

- Selecting a missing data option now adds a `DocumentReference` to the patient record moves to next state in pathway. The branch node will then read in the note and display details when it is expanded.
image
------
image
------
Screen Shot 2020-03-05 at 3 44 43 PM

- Accepting an action adds the recommended resource to the patient record and marks the node as in progress.

image

- Declining an action adds a DocumentReference with an attached note to the patient record and marks the node as in progress but declined.

Screen Shot 2020-03-05 at 2 43 07 PM copy

Still working on writing a note back when accepting/declining.


src/engine/output-results.ts
```diff
@@ -217,6 +230,7 @@ function nextState(
         status: 'status' in resource ? resource.status : 'unknown'
       };
     } else {
+      // TODO: possibly add search for a note here
```
dehall: Yeah I think you will have to look for Declines here

.eslintrc.json
```diff
@@ -24,7 +24,7 @@
     "@typescript-eslint/semi": "error",
     "brace-style": "off",
     "func-call-spacing": "off",
-    "max-len": ["warn", { "code": 100 }],
+    "max-len": ["warn", { "code": 100, "ignoreStrings": true }],
```
radamson, line 27: Instead of overriding the rule, break the string over multiple lines to maintain readability.

e.g.

```
const string = `my really long\
 string`;
```

src/components/App.tsx
```diff
@@ -15,17 +17,25 @@ import { EvaluatedPathway } from 'pathways-model';
...
-
+import { MockedFHIRClient } from 'utils/MockedFHIRClient';
+import { getHumanName } from 'utils/fhirUtils';
 interface AppProps {
   demo: boolean;
 }
 
 const App: FC<AppProps> = ({ demo }) => {
-  const [patientRecords, setPatientRecords] = useState<Array<fhir.DomainResource>>([]);
+  const [patientRecords, _setPatientRecords] = useState<fhir.DomainResource[]>([]);
```
dehall, line 27: Why start the name with an underscore? I thought that traditionally meant "unused" but I see it's used below (line 37 as of now)

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -154,7 +187,7 @@ function renderBranch(
       <ExpandedNodeField
         key="value"
         title="Value"
-        description={<MissingDataPopup values={values} />}
+        description={<MissingDataPopup values={values.filter((a, b) => values.indexOf(a) === b)} />}
```
dehall: Nitpick but it feels cleaner to do this "unique" test up above where the `values` field is first set. Also could we add a comment explaining that this is a unique test, it's not immediately obvious that's what it's doing.

.eslintrc.json
danlee1025: Is that what we want?  https://github.com/airbnb/javascript#strings--line-length

I was reading airbnb's javascript style guide suggesting not breaking it up so that it's more searchable.  I don't really care one way or the other but I really dislike string concat.

src/components/PatientProvider.tsx
```diff
@@ -3,7 +3,7 @@ import { useFHIRClient } from './FHIRClient';
 
 interface PatientProviderProps {
   children: ReactNode;
-  patient?: fhir.Patient;
+  patient?: fhir.Patient | null;
```
KeeyanGhoreshi, line 6: Should we really allow the patient to be null?  If the patient is null then I'd think the pathways app shouldn't even be rendering.

src/components/withConfirmationPopup/withConfirmationPopup.tsx
```diff
@@ -42,7 +42,14 @@ const PopupContent: FC<PopupContentProps> = ({ setOpen, onConfirm }) => {
     <div className={styles.popupContent}>
       <div>Are you sure?</div>
       <div>
-        <ActionButton size="small" type="accept" onClick={onConfirm} />
+        <ActionButton
+          size="small"
+          type="accept"
+          onClick={(): void => {
+            onConfirm?.();
```
KeeyanGhoreshi: Are we expecting to have action buttons without onConfirm functions?  This seems like it will just dump errors and make it hard to debug.

dehall: Agree - in the initial version when there was no implementation on any button it made sense to make it optional. At this point there shouldn't be any buttons that don't do anything.

.eslintrc.json
radamson: That works for me and I feel the 'painful to work with' and 'less searchable` arguments despite the reduced readability.

[updated the code style guide to reflect this new guidance.](https://github.com/mcode/pathways/wiki/Code-Style-Guide#strings)

dehall (changes requested):
In terms of overall functionality, this looks really good. 2 main things to point out:
1) Declines are written back as a note but not recognized as advancing the pathway -- the engine is going to have look for those as well, I think in the spot where there's currently a TODO
2) The format of the note for missing data should include more than just the name of the branch. If it could use the same format as the accept/decline notes that might be ideal.
I'll keep looking at the code as well

src/utils/fhirUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+      };
+    }
+    case 'MedicationRequest': {
+      const { medicationCodeableConcept } = pathwayResource as BasicMedicationRequestResource;
+      return {
+        ...resourceProperties,
+        medicationCodeableConcept
+      };
+    }
+    default: {
```
dehall, line 38: This should probably throw an error rather than just returning an empty object. With the latest changes do we have any other types in pathway actions?

danlee1025: I think those 2 are currently the other resource types.  I pushed a commit to throw an error here.

src/components/PatientProvider.tsx
danlee1025: It's null when the app isn't in demo mode and pulls the patient directly from the fhirclient.

src/components/withConfirmationPopup/withConfirmationPopup.tsx
danlee1025: Made `onConfirm` a required prop.

dehall:
One other thing I just noticed - when you select missing data and then re-open the applicable node, it still shows "missing data", for example on the demo patient choosing Tumor Size:
image

src/components/ExpandedNode/ExpandedNode.tsx
danlee1025: Addressed  this.

src/components/App.tsx
blangley28: Not sure if I wrote this specific change but I have been using this syntax to wrap the `setState` call in another method. Whenever `setPatientRecords` is run we also call `setEvaluatePath`. So the set method in the create here is `_setPatientRecords` and then on line 35 the actual `setPatientRecords` method is created:
```
const setPatientRecords = useCallback((value: fhir.DomainResource[]): void => {
    _setPatientRecords(value);
    setEvaluatePath(true);
  }, []);
```

dehall: Oh I missed the fact that where it was being called was called `setPatientRecords`. It's a tricky naming convention if you aren't familiar with it.

src/engine/output-results.ts
blangley28: Added search for notes on decline. Now reads note and marks the node as in progress but declined
Screen Shot 2020-03-05 at 2 44 11 PM

src/components/App.tsx
blangley28: Do you have any suggestion? I saw it elsewhere in the code base and decided to mimick it in a few places

blangley28:
Branch node which is documented by a DocumentReference reads in data from there. The `note` field is what is currently in the attachement.data, but will be updated to the new note structure (Dan is working on that now)
Screen Shot 2020-03-05 at 3 45 22 PM

src/utils/fhirUtils.ts
```diff
@@ -61,17 +61,25 @@ export function getHumanName(person: fhir.HumanName | fhir.HumanName[]): string
...
 
 export function createDocumentReference(
-  selected: string,
+  data: string,
+  labelOrCondition: string,
   patient: fhir.Patient
 ): fhir.DocumentReference {
   return {
     resourceType: 'DocumentReference',
+    id: btoa(labelOrCondition + Math.floor(Math.random() * 1000)),
```
radamson: This PR pulls in the `uuid` dependency for creating ids and it is used here: 
https://github.com/mcode/pathways/blob/6cc57a5e50e460f47c978d53a7601ea5ad634f23/src/utils/fhirUtils.ts#L2
https://github.com/mcode/pathways/blob/6cc57a5e50e460f47c978d53a7601ea5ad634f23/src/utils/fhirUtils.ts#L12.

Any reason for not using that library here?  Seems like we should otherwise.

blangley28: I think this was already updated

radamson: hmm, I don't see it updated...

danlee1025: Have you pulled the latest? It should be using uuid to generate the id now.

danlee1025:
I believe everything should be addressed.  Please re-review :D

dehall:
Tested with some demo patients using launch.smarthealthit.org and looks good! Can everyone who added features here update the description with the features they added?

KeeyanGhoreshi:
PR added note writeback, putting relevant information into an object and then converting it into a string to be stored in a documentReference resource, which it posts back to the EHR with the client.  Will write a note for accept, decline, and missing data info population.

src/components/NoteProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+interface NoteProviderProps {
+  children: ReactNode;
+  date: Date;
+  physician: string;
+}
+
+export const NoteContext = createContext<Note | null>(null);
+
+export const NoteProvider: FC<NoteProviderProps> = ({ children, date, physician }) => {
```
dehall: This NoteProvider feels a little weird to me -- it's not actually providing the Note per se, it's providing a couple pieces of data to the note like the date (which should also probably be fetched at the time of the click instead of at render, though in practice those probably won't be all that different) and the clinician/user TO the note generation process. Based on an offline conversation with @KeeyanGhoreshi I don't think it's worth totally breaking up at this point since it works, but could we at least rename it like to `NoteDataProvider` or something like that?

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -143,13 +188,48 @@ function renderBranch(
               }
             />
           );
+        break;
+      }
+      case 'DocumentReference': {
+        const documentReference = documentation.resource as fhir.DocumentReference;
+        const subject = documentReference.subject;
+        subject &&
```
dehall: I see this is following along with the style from above, but in general this approach of `object && thing-to-do-if-object-not-null()` is not something we should be doing in pure JS code. Ideally these should be `if(object) { thing-to-do-if-object-not-null() }`. Where we would use the `&&` approach is in JSX because you can't do `if`  blocks within JSX.

blangley28: I will go and update this throughout ExpandedNode.tsx

dehall (approved):

src/components/App.tsx
```diff
@@ -35,29 +45,35 @@ const App: FC<AppProps> = ({ demo }) => {
           scope: 'launch/patient openid profile'
         })
         .then(client => {
+          // TODO: MockedFHIRClient has not mocked out requests for resources yet
+          getPatientRecord(client).then((records: fhir.DomainResource[]) => {
+            // filters out values that are empty
+            // the server might return deleted
```
radamson, line 51: I don't have a problem with deleting these useless resources, but I don't think "deleted" is the right word for it because they do exist and are likely valid FHIR (depending on the resource/profile).

dehall: Note that this was just moved, not actually new. This was a change added a while back because we noticed some server was returning basically "empty" resources, i.e. they only had a resourceType, id, and meta. I think when we investigated the resources actually had been deleted but were still being returned in $everything for some reason.

src/components/App.tsx
```diff
@@ -35,29 +45,35 @@ const App: FC<AppProps> = ({ demo }) => {
...
-        // id, meta, and resourceType
-        const values = ['id', 'meta', 'resourceType'];
-        records = records.filter(resource => {
-          return !Object.keys(resource).every(value => values.includes(value));
-        });
-
-        setPatientRecords(records);
-      });
-    }
+    client?.user?.read().then((user: fhir.Practitioner) => {
```
radamson, line 71: No recommendation for change, but just a note:

We expect practitioners to be the primary user of this app, but this could return other types. e.g. If a patient was using the app then a `fhir.Patient` resource would get returned.  We could make this type more permissive by only requiring a name (maybe even the name is optional).

Again, I don't think we need a change here, but I could _maybe_ see a scenario where I'm a patient and I want to poke around the pathway I'm on to see what lies down the road.

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -57,10 +90,22 @@ const ExpandedNode: FC<ExpandedNodeProps> = ({
...
+              }}
+            />
           </div>
           <div className={styles.footer}>
-            <ConfirmedActionButton type="decline" size="large" />
+            <ConfirmedActionButton
+              type="decline"
+              size="large"
+              onConfirm={(): void => {
+                onConfirm('Declined', []);
```
radamson: I think it would look cleaner to make the argument optional instead of passing an empty array if we can.

https://github.com/mcode/pathways/blob/6df68acd6e6f281d67ce13e908ed864f75048bb8/src/components/ExpandedNode/ExpandedNode.tsx#L42

src/components/MissingDataPopup/MissingDataPopup.tsx
```diff
@@ -64,7 +73,20 @@ const PopupContent: FC<PopupContentProps> = ({ values, setOpen }) => {
...
+            size="small"
+            type="accept"
+            onClick={(): void => {
+              setOpen(false);
+              // Create DocumentReference with selected value and add to patient record
+              if (note) {
+                const noteString = createNoteContent(note, patientRecords, 'completed', selected);
+                const documentReference = createDocumentReference(noteString, selected, patient);
+                setPatientRecords([...patientRecords, documentReference]);
+                client?.create?.(documentReference);
```
radamson, line 86: I'm be a little worried about what happens here if the create operation fails.  Would we still see the documentreference in the patient record in the app even if this returns with a 4XX?

Everything seems to work so we don't have to change it here, but curious as to what would happen.

src/components/Navigation/Navigation.tsx
```diff
@@ -1,14 +1,13 @@
 import React, { FC } from 'react';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+// import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
radamson, line 2: remove commented code
```suggestion
```

src/components/Navigation/Navigation.tsx
```diff
@@ -1,14 +1,13 @@
...
 
 import PatientSnapshot from 'components/PatientSnapshot';
 import DropDown from 'components/DropDown';
-
 import styles from './Navigation.module.scss';
 import { EvaluatedPathway } from 'pathways-model';
 import { Option } from 'option';
 import { usePathwayContext } from 'components/PathwayProvider';
-import { faChevronLeft } from '@fortawesome/free-solid-svg-icons';
+// import { faChevronLeft } from '@fortawesome/free-solid-svg-icons';
```
radamson, line 10:
```suggestion
```

radamson (commented):
Looks good!

I left a couple comments for discussion and some minor change suggestions.

src/components/App.tsx
radamson: That makes sense and I think the description of "empty" more accurately describes it.

src/components/MissingDataPopup/MissingDataPopup.tsx
dehall: You can actually see it on a DSTU2 server -- for some reason it fails because there's an `id` included in the resource, even though it works fine on an R4 server. Yes we still see the DocumentReference in the patientrecords, even though it's not on the server. I don't want to hold this up for error handling though, let's make a backlog item for that.

src/components/Navigation/Navigation.tsx
dehall: Agree in general, though in this case Keeyan's other PR re-introduces the back button with conditional rendering rather than just commenting it out, so it's not a huge deal either way

src/components/ExpandedNode/ExpandedNode.tsx
danlee1025: Agreed, I feel like when I tried this I was getting Typescript errors.  I made the change.  Do you know why on line 54 Typescript  doesn't like `action?.length > 0`... I had to make it `action && action.length > 0`

radamson: Seems upset because you can't compare use compare 0 with undefined using `>`.

Found pretty much the exact issue here: https://github.com/microsoft/TypeScript/issues/34593, but I like your solution better than the workaround proposed in the issue.

radamson (approved):

PR #22, opened 2020-03-05
Node header

KeeyanGhoreshi:
removes back button, makes your mouse a pointer on node headers.

### DIFF
--- a/src/styles/_nodes.scss
+++ b/src/styles/_nodes.scss
@@ -20,6 +20,7 @@
 }
 
 .nodeTitle {
+  cursor: pointer;
   padding: 15px;
   /* use the entire width of its parent 
      so we can align something to the right */
--- a/src/styles/_nodes.scss
+++ b/src/styles/_nodes.scss
@@ -20,7 +20,6 @@
 }
 
 .nodeTitle {
-  cursor: pointer;
   padding: 15px;
   /* use the entire width of its parent 
      so we can align something to the right */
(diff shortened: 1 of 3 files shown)


dehall:
So the back button change looks good. Tested with adding `onBack={() => window.history.back()}` to the Navigation item in App.tsx and it worked as expected.
About the node header pointer cursor, it's also a pointer on the Explore Pathways screen where it's not actually clickable. It should probably only appear clickable on the main pathways view. Maybe with a new className that only gets set on the Node if the onClickHandler isn't null.
Last thing is this needs a rebase.

Also I deleted the Jira issue for this when you said the entire node title was already clickable. I forgot about the pointer cursor thing. I'll add that back now.

danlee1025:
Can this branch get a rebase?  I'll look at it then.

KeeyanGhoreshi:
rebased

danlee1025:
This looks pretty good, once the cursor pointer is removed in the explore pathways screen I'll approve.

src/styles/_nodes.scss
```diff
@@ -29,6 +29,10 @@
   justify-content: flex-end;
 }
 
+.clickable {
+    cursor: pointer;
```
dehall:
```suggestion
  cursor: pointer;
```
Code style is this should be indented 2 spaces

dehall (approved):

radamson (approved):

PR #23, opened 2020-03-06
PathwaysList order by matched elements

blangley28:
Modified the data model a little and tests are failing so marking as WIP while I fix those.

Orders the pathways in the list by the number of elements matched (descending). Per the task description this waits until all the criteria have been evaluated to display the pathway list. These results are not cached anywhere so when moving back to the explore screen from viewing a pathway there is another delay to reload the criteria.

Screen Shot 2020-03-06 at 4 22 27 PM

## Criteria Evaluation Result
 - `pathwayName`: string - the name of the pathway the criteria results are associated to
 - `matches`: number - total count of matched criteria result items
 - `criteriaResultItems`: CriteriaResultItem[] - list of the each evaluated criteria item

## Criteria Evaluation Result Item
 - `elementName`: string - friendly name of the mCODE element
 - `expected`: string - friendly human readable description of the expected value
 - `actual`: string - result from the patient's record indicating the actual value of the selected mCODE element
 - `match`: boolean - whether or not `expected` matches `actual`. Note that in most cases strict string/number equality checking is insufficient


src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -36,19 +37,43 @@ const PathwaysList: FC<PathwaysListProps> = ({
...
+              return (
+                <PathwaysListElement
+                  evaluatedPathway={evaluatedPathway}
+                  callback={callback}
+                  resources={resources}
+                  criteria={c}
+                  key={evaluatedPathway.pathway.name}
+                />
+              );
+            else return <div></div>;
```
dehall: Could we put something here as a loading indicator? Nothing too crazy, just something that says "yes the system is working to load a pathway" when `criteria` is null. It will be especially nice on the demo endpoint when things move slowly.

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -19,6 +19,7 @@ import {
 
 interface PathwaysListElementProps {
   evaluatedPathway: EvaluatedPathway;
+  criteria: CriteriaResult | undefined;
```
dehall: Not a big deal but this can be changed to 
```suggestion
  criteria?: CriteriaResult;
```

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -36,19 +37,43 @@ const PathwaysList: FC<PathwaysListProps> = ({
...
+    const patient = {
+      resourceType: 'Bundle',
+      entry: resources.map((r: fhir.Resource) => ({ resource: r }))
+    };
+
+    // Evaluate pathway criteria for each pathway
+    const criteriaPromises: Promise<CriteriaResult>[] = [];
+    evaluatedPathways.forEach(pathway => {
+      criteriaPromises.push(evaluatePathwayCriteria(patient, pathway.pathway));
+    });
```
dehall: Also not a big deal but this kind of thing is a good use for the `map` function:
```suggestion
    const criteriaPromises = evaluatedPathways.map(pathway => evaluatePathwayCriteria(patient, pathway.pathway));
```

src/types/pathways-model.d.ts
```diff
@@ -91,6 +91,12 @@ declare module 'pathways-model' {
     match: boolean; // in case expected !== actual but they are still a match
   }
 
+  export interface CriteriaResult {
```
dehall, line 94: Don't forget we'll have to update the data model page on the wiki as well. Since the wiki isn't versioned alongside the code, could you post a comment or edit the description of this PR with the relevant piece of the wiki so we can just copy & paste when merging?

src/components/PathwaysList/PathwaysList.tsx
blangley28: So I can certainly add that functionality but this else is for when the app cannot find a matching pathway for the criteria result. It should never happen but I had to put the else in

src/types/pathways-model.d.ts
blangley28: Updated the description of the PR

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -36,19 +37,42 @@ const PathwaysList: FC<PathwaysListProps> = ({
   service,
   resources
 }) => {
+  const [criteria, setCriteria] = useState<CriteriaResult[] | null>(null);
+
+  if (!criteria && evaluatedPathways.length > 0 && resources != null && resources.length > 0) {
```
radamson:
```suggestion
  if (!criteria && evaluatedPathways.length > 0 && resources && resources.length > 0) {
```
don't need to check against null, it will be considered true if it's not falsy.

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -112,7 +127,7 @@ const PathwaysListElement: FC<PathwaysListElementProps> = ({
         <div className={styles.expand}>
           <FontAwesomeIcon icon={chevron} />
         </div>
-        <div className={styles.numElements}>{criteria && criteria.filter(c => c.match).length}</div>
+        <div className={styles.numElements}>{criteria && criteria.matches}</div>
```
radamson:
```suggestion
        <div className={styles.numElements}>{criteria?.matches}</div>
```

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -127,7 +142,7 @@ const PathwaysListElement: FC<PathwaysListElementProps> = ({
                   <th>patient elements</th>
                 </tr>
                 {criteria &&
-                  criteria.map(c => (
+                  criteria.criteriaResultItems.map(c => (
```
radamson: for some reason I can't get multiline suggestion to work, but:

`criteria?.criteriaResultItems?.map`

src/components/PathwaysList/__tests__/PathwaysList.test.tsx
```diff
@@ -28,18 +51,23 @@ describe('<PathwaysList />', () => {
...
           return;
         }}
         service={loadedService}
-        resources={[]}
+        resources={resources}
       />
     );
-    expect(getAllByText(/test./)).toHaveLength(3);
+    if (result) expect(result.getAllByText(/test./)).toHaveLength(3);
+    else fail();
```
radamson:
```suggestion
    expect(result?.getAllByText(/test./)).toHaveLength(3);
```

radamson (commented):
Only had a few minor syntax change suggestions.

Also I think this needs a rebase onto master.  ~The current version of master has the missing data implementation that would make it easier to see this functionality in action.~ oops, I forgot about the demo endpoint.

src/components/PathwaysList/PathwaysList.tsx
blangley28: Updated here and line 61/62

dehall (approved):

radamson (approved):

PR #24, opened 2020-03-09
Add R4 FHIR types

KeeyanGhoreshi:
Basically creates a module `fhir-objects` that abstracts the STU3 and R4 types to a single type that is a union of the two.  Anywhere that used to use the fhir types has been replaced with one of these union types.  

In basically only one circumstance I found did the type difference actually matter, since STU3 and R4 are very similar anyway.  The union types can be used as standalone types for most purposes, aka you don't have to cast to a specific, versioned type and write two different functions if you don't need a property that is different between the two versions.

### DIFF
--- a/src/types/cql-exec.fhir.d.ts
+++ b/src/types/cql-exec.fhir.d.ts
@@ -2,8 +2,8 @@ declare module 'cql-exec-fhir' {
   import { Bundle } from 'fhir-objects';
   export const PatientSource = {
     FHIRv400: class {
-      constructor(patient: Bundle); // TODO: FHIR type
-      loadBundles(bundle: Bundle): void; // FHIR type here too
+      constructor(patient: Bundle);
+      loadBundles(bundle: Bundle): void;
     }
   };
 }
--- a/src/types/cql-exec.fhir.d.ts
+++ b/src/types/cql-exec.fhir.d.ts
@@ -1,8 +1,9 @@
 declare module 'cql-exec-fhir' {
+  import { Bundle } from 'fhir-objects';
   export const PatientSource = {
     FHIRv400: class {
-      constructor(patient: object); // TODO: FHIR type
-      loadBundles(bundle: object): void; // FHIR type here too
+      constructor(patient: Bundle); // TODO: FHIR type
+      loadBundles(bundle: Bundle): void; // FHIR type here too
     }
   };
 }
--- a/src/types/fhir-objects.d.ts
+++ b/src/types/fhir-objects.d.ts
@@ -18,5 +18,4 @@ declare module 'fhir-objects' {
   export type Bundle = fhir.Bundle | R4.IBundle;
   export type ServiceRequest = fhir.ProcedureRequest | R4.IServiceRequest;
   export type MedicationRequest = fhir.MedicationRequest | R4.IMedicationRequest;
-  export type Request = ServiceRequest | MedicationRequest;
 }
--- a/src/types/fhir-objects.d.ts
+++ b/src/types/fhir-objects.d.ts
@@ -15,4 +15,8 @@ declare module 'fhir-objects' {
   export type Practitioner = fhir.Practitioner | R4.IPractitioner;
   export type Procedure = fhir.Procedure | R4.IProcedure;
   export type Resource = fhir.Resource | ResourceR4;
+  export type Bundle = fhir.Bundle | R4.IBundle;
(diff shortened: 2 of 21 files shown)


dehall:
@KeeyanGhoreshi and reviewers -- please also do a quick scan of the code for any spots where we currently use `object` or `any` or something like that in place of actual FHIR types. I know in at least a couple places we have some TODOs to that effect

blangley28:
I identified a few places we many need to update `object` (it does not look like `any` is being used anywhere) to use FHIR resources

1. `Graph.tsx` line 107 - `resources` is type `object[]`
2. `evaluate-patient.ts` line 22 - `patient` is type `object`
3. `evaluate-patient.ts` line 24 - `resources` is type `object[]`
4. `evaluate-patient.ts` line 39 - `patient` is type `object`
5. `evaluate-patient.ts` line 54 - `patient` is type `object`
6. `output-results.ts` line 36 - `resources` is type `object[]`
7. `cql-exec.fhir.d.ts` line 4 - `patient` is type `object`
8. `cql-exec.fhir.d.ts` line 5 - `bundle` is type `object`

In `pathways-model.d.ts` there is also a comment for updating `BasicActionResource` and `BasicMedicationRequestResource` to use FHIR types

src/types/fhir-objects.d.ts
```diff
@@ -0,0 +1,22 @@
...
+  export type Identifier = fhir.Identifier | R4.IIdentifier;
+  export type Observation = fhir.Observation | R4.IObservation;
+  export type Patient = fhir.Patient | R4.IPatient;
+  export type Practitioner = fhir.Practitioner | R4.IPractitioner;
+  export type Procedure = fhir.Procedure | R4.IProcedure;
+  export type Resource = fhir.Resource | ResourceR4;
+  export type Bundle = fhir.Bundle | R4.IBundle;
+  export type ServiceRequest = fhir.ProcedureRequest | R4.IServiceRequest;
+  export type MedicationRequest = fhir.MedicationRequest | R4.IMedicationRequest;
+  export type Request = ServiceRequest | MedicationRequest;
```
dehall: Since this Request type is our own use-case based aggregation of otherwise independent types, can we somehow identify it as such? Not a huge deal but I would consider either renaming this, moving this elsewhere, or not making it a formal "type" at all.

src/types/cql-exec.fhir.d.ts
```diff
@@ -1,8 +1,9 @@
 declare module 'cql-exec-fhir' {
+  import { Bundle } from 'fhir-objects';
   export const PatientSource = {
     FHIRv400: class {
-      constructor(patient: object); // TODO: FHIR type
-      loadBundles(bundle: object): void; // FHIR type here too
+      constructor(patient: Bundle); // TODO: FHIR type
+      loadBundles(bundle: Bundle): void; // FHIR type here too
```
dehall: we can remove these TODOs now

dehall (approved):
Some minor suggestions below but overall looks good

dehall:
Now this is failing travis on `yarn build`. Failing locally for me too but with a different error:
```
/Users/dehall/pathways/src/components/ExpandedNode/ExpandedNode.tsx
TypeScript error in /Users/dehall/pathways/src/components/ExpandedNode/ExpandedNode.tsx(275,17):
Property 'code' does not exist on type 'never'.  TS2339

    273 |   const coding = isMedicationRequest(resource)
    274 |     ? resource?.medicationCodeableConcept?.coding
  > 275 |     : resource?.code?.coding;
        |                 ^
    276 |
    277 |   const returnElements = [
    278 |     <ExpandedNodeField
```

KeeyanGhoreshi:
I deleted node_modules and reinstalled and that error went away for me.

KeeyanGhoreshi:
Also the error the travis build is failing on:

`Property 'type' is missing in type '{ resourceType: string; entry: { resource: fhir.Resource; }[]; }' but required in type 'Bundle'.  TS2345`

Is explicitly fixed in my latest commit.  I added a type to the object.

dehall (approved):
Restarted the travis build and it worked fine. Deleted node_modules and reinstalled and worked locally. Looks good. Needs a rebase though 😝

radamson (approved):
Not sure I'm sold on why we need to union STU3 and R4, if anything I'd expect DSTU2, but we should watch out and be careful here because there are some differences that could be significant.

one thing that comes to my mind is.

`ReferralRequest` & `ProcedureRequest` were renamed to `ServiceRequest`

PR #26, opened 2020-03-13
Added Themes

noranda:
- Added a menu to the header which has a "Toggle Theme" option. This will change between the default theme (current) and the projector theme (darker grays that won't wash out when projected). 
- Added [`material-ui`](https://material-ui.com/) to handle theming. We may want to consider moving all style sheets over to use material-ui for consistency and to make future theming easy.
- Added [`clsx`](https://www.npmjs.com/package/clsx) to better handle creating class names.

To test: click on gear icon in header and select "Toggle Theme".


radamson:
I think it would be cool to use JSS (which is what [material-ui uses under the hood](https://material-ui.com/styles/basics/#why-use-material-uis-styling-solution)).  Agree we might want to eventually move everything over at some point if we go this route and everyone agrees.

If we're only going to use material-ui for the theme's though we could just use JSS directly and get the same results.  [The material-ui theming functions appear to just be the JSS ones really for the most part.](https://cssinjs.org/react-jss/?v=v10.0.4#theming)

If we decide to use material-ui components themselves, which I also think could be good and it seems like you've included a few, we should make sure the whole team knows to start grabbing stuff from there instead of building stuff from scratch when possible.

Excited to test this out and I think these will be great additions, as per usual 😉 .

noranda:
@radamson Actually I'm already using Material UI for the menu and think it'll be useful for the patient record drawer as well. I'm all for it. :)

dehall:
Tested and looks good on my screen at least. Wish we had someone still in the office to try this out on the projectors  😞

dehall (approved):

src/components/Node/Node.tsx
```diff
@@ -42,15 +57,19 @@ const Node: FC<NodeProps & { ref: Ref<HTMLDivElement> }> = forwardRef<HTMLDivEle
     },
     ref
   ) => {
+    const classes = useStyles();
```
radamson, line 60: I know material-ui's examples use the variable name `classes` for styles, but our style guide suggests the name `styles`.  `classes` is a bit more of an overloaded term and it's not immediately obvious that it's talking about _style_ classes specifically.

This is also the case in: `Navigation`, `PathwaysList`, and `PatientSnapshot`

radamson (commented):
Looks great 👍 .

One thing I noticed is that `src/styles/theme.js` is javascript instead of TypeScript.  At this point we're committed to using TypeScript so we should convert this over.  Happy to help with that if desired.

src/components/Node/Node.tsx
noranda: Agree with this, but since we're already using `styles`, we have a namespace issue. If we decide to merge all the styles together using material UI, we can have everything under `styles`. I didn't merge the stylesheets for these files for consistency, but we should revisit this and consider making a task to do so across the project.

radamson (approved):

PR #29, opened 2020-03-17
Add ELM to pathway data model

dehall:
Adds preconverted ELM to the pathway data model, which, if present, is used to evaluate the patient against the pathway. As you might imagine this makes the app a lot faster and removes the runtime dependency on a CQL-to-ELM webservice.

For now this field is optional, so there is still the code flow where if necessary the pathway CQL is extracted, converted to ELM, and evaluated. Optional primarily because we don't have a good way to manually convert the pathway CQL to ELM. Once we have a pathway builder all of that should be taken care of for us, and we can make the field required.
The data model adds an `elm` field to the pathway object, which contains two subfields: 
 - `navigational`: which contains the converted ELM for the navigational elements within the pathway, i.e. branch nodes and recommendation nodes
 - `criteria`: which contains the converted ELM for the criteria preconditions for the pathway
I'm not sure if there's an easy way to combine those -- we can make that a separate task in the backlog if it's desirable.

@rdingwell  @zlister  heads up for the data model changes. Let me know if any concerns


src/types/pathways-model.d.ts
```diff
@@ -8,6 +8,13 @@ declare module 'pathways-model' {
     states: {
       [key: string]: GuidanceState | BranchState | State;
     };
+    elm?: PathwayELM;
```
KeeyanGhoreshi, line 11: I feel like this should remain optional.  It seems presumptive to assume that the app is only allowed to work when provided ELM, when we have the ability to convert CQL already.

src/engine/evaluate-patient.ts
```diff
@@ -67,21 +79,30 @@ function processCQLCommon(patient: Bundle, cql: string): Promise<PatientData> {
...
-      return patientData;
-    });
+/**
+ * Common logic to execute the given ELM against the given Patient.
+ * @param patient - Patient's record as FHIR data
+ * @param elm - resulting ELM from converting the CQL in a pathway
+ * @return the raw, unprocessed patientResults
+ *         derived from executing the ELM against the given patient
+ */
+function processELMCommon(patient: Bundle, elm: object): PatientData {
```
KeeyanGhoreshi: Also this probably should have been addressed in my R4 typing PR, but having `patient` be the variable name for what is actually a bundle of the patient's record is pretty confusing.  It should be updated at some point.

src/types/pathways-model.d.ts
dehall: The idea is that once we have a pathway builder, it should have the ability to convert CQL to ELM directly, so at that point there won't be any need to create pathways manually and hand write CQL, and there shouldn't be any instances where the ELM isn't present.

src/engine/evaluate-patient.ts
dehall: What do you want to call it? From the type information we know it's a Bundle, and it's not supposed to be any arbitrary bundle, but a single patient's entire record.

KeeyanGhoreshi: I'd suggest renaming to patientRecord.

dehall: Renamed to patientRecord

src/engine/evaluate-patient.ts
```diff
@@ -13,45 +13,57 @@ function instanceOfElmObject(object: object): object is ElmObject {
...
 export function evaluatePatientOnPathway(
-  patient: Bundle,
+  patientRecord: Bundle,
   pathway: Pathway,
   resources: DomainResource[]
 ): Promise<PathwayResults> {
-  return extractNavigationCQL(pathway)
-    .then(cql => processCQLCommon(patient, cql))
-    .then(patientData => pathwayData(pathway, patientData, resources));
+  if (pathway.elm && pathway.elm.navigational) {
```
radamson:
```suggestion
  if (pathway?.elm?.navigational) {
```

src/engine/evaluate-patient.ts
```diff
@@ -67,21 +79,30 @@ function processCQLCommon(patient: Bundle, cql: string): Promise<PatientData> {
...
-      return patientData;
-    });
+/**
+ * Common logic to execute the given ELM against the given Patient.
+ * @param patientRecord - Patient's record as FHIR data
+ * @param elm - resulting ELM from converting the CQL in a pathway
+ * @return the raw, unprocessed patientResults
+ *         derived from executing the ELM against the given patient
+ */
+function processELMCommon(patientRecord: Bundle, elm: object): PatientData {
```
radamson: because `processCQLCommon` returns a Promise, you can save a few lines of code in other places by returning a promise here too (plus maybe avoid any astonishment from users who see the similar names).

```suggestion
function processELMCommon(patientRecord: Bundle, elm: object): Promise<PatientData> {
```

e.g. the `evaluatePathwayCriteria` method could be condensed to 4 lines instead of 8 and be a hair DRYer:


```typescript
export function evaluatePathwayCriteria(
  patientRecord: Bundle,
  pathway: Pathway
): Promise<CriteriaResult> {
  const patientData: Promise<PatientData> = pathway?.elm?.criteria
    ? processELMCommon(patientRecord, pathway.elm.criteria)
    : extractCriteriaCQL(pathway).then(cql => processCQLCommon(patientRecord, cql));
  return patientData.then(patientData => criteriaData(pathway, patientData));
}
```

Similarly for `evaluatePatientoOnPathway`.

radamson (commented):
LGTM, although I'm not a CQL or ELM expert.  I left a super small comment regarding changing the return type on `processELMCommon` to be match the return type of `processCQLCommon` because it can help DRY out the code in other places.

radamson (approved):
💯

danlee1025 (approved):
Looks good!

PR #30, opened 2020-03-23
Handle small window size

danlee1025:
- Adds scrollbar to div containing graph when content overflows
- Shifts nodes and edges to the right if content overflows to the left(could not figure out how to make scrollbar appear when content overflowed to the left)

image

### DIFF
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -80,6 +80,29 @@ const Graph: FC<GraphProps> = ({
       : 0;
   }, [nodeCoordinates]);
 
+  // If a node has a negative x value, shift nodes and edges to the right by that value
+  const minX =
+    nodeCoordinates !== undefined
+      ? Object.values(nodeCoordinates)
+          .map(x => x.x + windowWidth / 2)
+          .reduce((a, b) => Math.min(a, b))
+      : 0;
+
+  if (minX < 0) {
+    const toAdd = minX * -1;
+    Object.keys(nodeCoordinates).forEach(key => {
+      const node = nodeCoordinates[key];
+      node.x += toAdd;
+    });
+
+    Object.keys(edges).forEach(key => {
+      const edge = edges[key];
+
+      edge.points.forEach(p => (p.x += toAdd));
+      if (edge.label) edge.label.x += toAdd;
+    });
+  }
+
   const initialExpandedState = useMemo(() => {
     return Object.keys(layout).reduce((acc: { [key: string]: boolean }, curr: string) => {
       acc[curr] = false;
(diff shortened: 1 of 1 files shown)


blangley28:
Few comments about functionality:
1. The horizontal scroll works in the explore screen but the vertical one doesn't. 
2. I think there should be some horizontal padding on the graph. When scrolling over the right most node only just comes into view. (See picture below)
3. It still looks like the "N0" label in the picture below is being cut off. I'm not sure if this is part of this task as it looks like the graph frame is large enough.

Screen Shot 2020-03-24 at 11 37 44 AM

src/components/Graph/Graph.tsx
```diff
@@ -80,6 +80,29 @@ const Graph: FC<GraphProps> = ({
       : 0;
   }, [nodeCoordinates]);
 
+  // If a node has a negative x value, shift nodes and edges to the right by that value
+  const minX =
```
blangley28, line 84: If there is left overflow this shifts everything to the right? This would mean the start node is no longer in the center. Is there a way to set the default scroll to make the start node in the center when first loading the page?

danlee1025:
2 and 3 should be addressed, still looking into 1.

danlee1025:
For 1, we define a height for the div that contains the graph so it doesn't produce a vertical scrollbar since the height we state should fit all of the content. If we don't define a height then we would also get a vertical scrollbar in the div and window.
image

@blangley28 Do you think we should define a separate height for the graph when in the explore pathways screen?

blangley28 (approved):

src/components/Graph/Graph.tsx
danlee1025: Well if it shifts everything to the right the start node would still be center of the graph, I could look into setting the default scroll, is the start node off screen for you right now?

src/components/Graph/Graph.tsx
```diff
@@ -139,18 +162,27 @@ const Graph: FC<GraphProps> = ({
...
   // maxWidth finds the edge label that is farthest to the right
   const maxWidth: number =
-    edges !== undefined
+    (edges !== undefined
       ? Object.values(edges)
           .map(e => e.label)
           .map(l => (l ? l.x + l.text.length * 10 + windowWidth / 2 : 0))
           .reduce((a, b) => Math.max(a, b), 0)
-      : windowWidth;
+      : windowWidth) + 5;
```
dehall: Minor point but could you add a comment explaining this + 5?

dehall (approved):
Overall looks great, confirmed it works on Edge/IE11 as well

dehall:
Not sure what happened to travis. Seems like it ran for the latest commit ( https://travis-ci.org/github/mcode/pathways/builds/667252101 ) but for some reason it didn't connect to GitHub correctly. I'm merging it in anyway

PR #31, opened 2020-03-24
Support Multiple Current States

blangley28:
The Neoadjuvant pathway shows an example of multiple options. The pathway now identifies all of them as the current node and expands them.
Screen Shot 2020-03-24 at 11 31 35 AM

### DIFF
--- a/public/static/pathways/neoadjuvant_pathway.json
+++ b/public/static/pathways/neoadjuvant_pathway.json
@@ -125,7 +125,7 @@
             "description": "Paclitaxel request",
             "resource": {
               "resourceType": "MedicationRequest",
-              "code": {
+              "medicationCodeableConcept": {
                 "coding": [
                   {
                     "system": "http://www.nlm.nih.gov/research/umls/rxnorm",
@@ -142,7 +142,7 @@
             "description": "Trastuzumab request",
             "resource": {
               "resourceType": "MedicationRequest",
-              "code": {
+              "medicationCodeableConcept": {
                 "coding": [
                   {
                     "system": "http://www.nlm.nih.gov/research/umls/rxnorm",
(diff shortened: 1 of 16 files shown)


dehall:
I just tested with the demo patient, and from the neoadjuvant pathway I selected T > 2 then accepted the TCHP state (the first one on the left). There is the bug with the fhir visualizers but for now we can just comment out all the patient visualizers. What I got looked like this, I would have expected the fact that I accepted one to un-highlight the other 2 in terms of being actionable. Also it's showing an (X) like it was declined instead of being accepted. The arrow is highlighted correctly though
Screen Shot 2020-03-26 at 7 30 32 AM

blangley28:
Fixed issue where remaining options remain marked as current state after selecting an option. Also fixed the bug where accepting the chemotherapy options in the neoadjuvant parkway mark it as declined (pathway + CQL errors). Also fixed bug where the last arrow was not highlighted.

To test remove `<PatientRecord />` on line 130 of `App.tsx`

Screen Shot 2020-03-27 at 8 31 00 AM
Screen Shot 2020-03-27 at 8 31 10 AM
Note: accepting the chemotherapy regimen in this pathway moves on to the next state because the action here is a MedicationRequest and not a ServiceRequest. The engine handles these differently. Created [PATHWAYS-216](https://jira.mitre.org/browse/PATHWAYS-216) to address it

src/components/Graph/Graph.tsx
```diff
@@ -35,13 +41,13 @@ const Graph: FC<GraphProps> = ({
...
-  const [path, _setPath] = useState<string[]>(
-    evaluatedPathway.pathwayResults ? evaluatedPathway.pathwayResults.path : []
+  const [pathwayResults, _setPathwayResults] = useState<PathwayResults | null>(
+    evaluatedPathway.pathwayResults
   );
 
   const setPath = useCallback(
     (value: PathwayResults): void => {
-      _setPath(value.path);
+      _setPathwayResults(value);
```
KeeyanGhoreshi: Aren't we setting the pathway in two places with this?  Once with this `_setPathwayResults`, and again in `updateEvaluatedPathways`.  The  pathway gets passed to the Graph as a prop already, so updating it in the parent should be enough.

blangley28: Good catch. Removed the pathwayResults state and updated all uses to use `evaluatedPathway.pathwayResults`

KeeyanGhoreshi:
Looks good, will approve with rebase

blangley28:
Rebased

KeeyanGhoreshi (approved):

KeeyanGhoreshi:
my approval does not count

public/static/pathways/neoadjuvant_pathway.json
```diff
@@ -161,15 +161,32 @@
       "ChemotherapyTCHP": {
         "label": "Chemotherapy TCHP",
         "action": [
+          {
+            "type": "create",
+            "description": "Trastuzumab Medication request",
```
dehall: Why does this action say description: "Trastuzumab" but the code below says Doxorubicin?

dehall (commented):
Overall looks great. Just one question on the pathway

public/static/pathways/neoadjuvant_pathway.json
blangley28: Probably because I copied and pasted then changed the code but forgot to change the description. Updated.

dehall (approved):

dehall:
Don't forget to update the data model on the wiki

blangley28:
Updated data model on the wiki

PR #32, opened 2020-03-25
Support multiple demo endpoints

mdube7:
Added support for an id parameter after the demo to use different files. for example:
http://localhost:3000/#/demo/Maureen
http://localhost:3000/#/demo/Maryam

To add more demo records, create the new record in `pathways\public\static\demoData'


src/components/App.tsx
```diff
@@ -134,12 +134,26 @@ const App: FC<AppProps> = ({ demo }) => {
     );
   };
 
+  function getDemoId(): number {
```
dehall: This works, but I'm wondering if there's an easier/cleaner way to do it in the router in index.js, something like a urlParam: https://reacttraining.com/react-router/web/example/url-params .

src/fixtures/DemoData.tsx
```diff
@@ -0,0 +1,4 @@
+import maureenDemo from 'fixtures/MaureenMcodeDemoPatientRecords.json';
+import maureenDemo2 from 'fixtures/MaureenMcodeDemoPatientRecords2.json';
+
+export const DemoRecords = [maureenDemo, maureenDemo2];
```
dehall: Again this definitely works, but there's the downside of we have to make code changes every time we want to add a new demo file. Maybe it's not possible with the files being in the `src` folder, but could you look into doing a `fetch` for patient records under the `public/static` folder, similar to what we do for CQL libraries? We could enforce a filename pattern or just say file `maureenDemo.json` gets read when you go to `http://pathways/#/demo/maureenDemo`

blangley28: Up for discussion since this is the only export in the module but `DemoRecords` is a variable so I think it should be `demoRecords`. For reference the code style guide is here: https://github.com/mcode/pathways/wiki/Code-Style-Guide

src/components/App.tsx
```diff
@@ -134,12 +134,26 @@ const App: FC<AppProps> = ({ demo }) => {
...
+    // parse the url for the id of the patient to use
+    const url = window.document.location.href;
+    const params = url.split('?');
+    const urlParams = new URLSearchParams(params[1]);
```
blangley28: You don't need to use split here. The following should work
`const urlParams = new URLSearchParams(window.location.search);`

src/components/App.tsx
mdube7: refactored to use params

src/fixtures/DemoData.tsx
mdube7: refactored to fetch records from public/static

src/index.js
```diff
@@ -11,8 +11,17 @@ const rootElement = document.getElementById('root');
 ReactDOM.render(
   <Router>
     <Switch>
-      <Route exact path="/" render={() => <App demo={false} />}></Route>
-      <Route path="/demo" render={() => <App demo={true} />}></Route>
+      <Route exact path="/" render={() => <App demo={false} demoData={''} />}></Route>
```
blangley28: `demoData` should probably just be made optional instead of passing in an empty string

src/components/App.tsx
```diff
@@ -18,15 +18,16 @@ import ThemeProvider from './ThemeProvider';
 import { EvaluatedPathway } from 'pathways-model';
 import useGetPathwaysService from './PathwaysService/PathwaysService';
 import FHIR from 'fhirclient';
-import demoRecords from 'fixtures/MaureenMcodeDemoPatientRecords.json';
 import { MockedFHIRClient } from 'utils/MockedFHIRClient';
 import { getHumanName } from 'utils/fhirUtils';
 import { Patient, DomainResource, Practitioner } from 'fhir-objects';
 interface AppProps {
   demo: boolean;
+  demoId: string;
```
blangley28: Make this optional 
```suggestion
  demoId?: string;
```

src/components/PatientProvider.tsx
```diff
@@ -1,27 +1,29 @@
...
 }
 
-export const PatientContext = createContext<Patient | null>(null);
-
-export const PatientProvider: FC<PatientProviderProps> = ({ children, patient }) => {
-  const client = useFHIRClient();
-  const [currentPatient, setCurrentPatient] = useState<Patient | null>(patient || null);
+interface PatientContextInterface {
+  patient: Patient | null;
+  setPatient: Function;
```
blangley28, line 10: Maybe I am missing something but it seems like `setPatient` is never used in this context. The only time it is called is in the `App` component but that is using the `setPatient` state method created in `App` not from the context. Is there a reason this changed?

public/config.js
```diff
@@ -2,5 +2,6 @@ CONFIG = {
   mapper: null, // mapper name from fhir-mapper -- allows mapping of incoming FHIR resources
   cqlToElmWebserviceUrl: 'http://moonshot-dev.mitre.org:8080/cql/translator',
   pathwaysService: 'http://pathways.mitre.org:3002/pathways/',
-  demoPathwaysService: './static/pathways/'
+  demoPathwaysService: './static/pathways/',
+  demoPatientUrl: './static/demoData/'
```
dehall: Can we call this `demoPatients` to be more specific?

src/components/App.tsx
```diff
@@ -18,17 +18,17 @@ import ThemeProvider from './ThemeProvider';
...
 import FHIR from 'fhirclient';
-import demoRecords from 'fixtures/MaureenMcodeDemoPatientRecords.json';
 import { MockedFHIRClient } from 'utils/MockedFHIRClient';
 import { getHumanName } from 'utils/fhirUtils';
 import { DomainResource, Practitioner } from 'fhir-objects';
 import styles from './App.module.scss';
-
 interface AppProps {
   demo: boolean;
+  demoId: string;
```
dehall: We could theoretically condense these into one field... `demoId` being an optional string, if it's null then use the FHIR client. Not super high priority

src/components/App.tsx
```diff
@@ -65,13 +65,21 @@ const App: FC<AppProps> = ({ demo }) => {
...
         });
     } else {
       setClient(new MockedFHIRClient());
-      setPatientRecords(demoRecords);
+      const url = config.get('demoPatientUrl') + demoId + '.json';
+      fetch(url)
+        .then(data => data.json())
+        .then(result => {
+          const resultPatient = result.find((r: DomainResource) => r.resourceType === 'Patient');
+          setPatientRecords(result);
```
dehall, line 77: For future consideration only, not here, I wonder if it makes sense to combine the PatientRecordsProvider and PatientProvider

dehall (commented):
Overall looks good. I feel like there are probably ways it could be cleaned up a little further, but I don't see anything specifically.
One thought: we should probably pick a naming convention for demo patient files. Originally I thought we would just use numbers but maybe someone has a better idea. Since we have `/demo/` in the URL I don't think we need `Demo` in the filename too though.

Also could you replace MaureenDemo2 with this patient record? It's more different and is actually FHIR R4.
[maryam.json.txt](https://github.com/mcode/pathways/files/4429247/maryam.json.txt)

public/config.js
mdube7: Updated name

src/components/App.tsx
mdube7: removed demo

src/components/App.tsx
```diff
@@ -84,7 +91,7 @@ const App: FC<AppProps> = ({ demo }) => {
   }, [client]);
 
   const service = useGetPathwaysService(
-    config.get(demo ? 'demoPathwaysService' : 'pathwaysService')
+    config.get(demoId === null ? 'demoPathwaysService' : 'pathwaysService')
```
blangley28: This is inconsistent with the prior behavior - `demo === true` should transform to `demoId !== null`
```suggestion
    config.get(demoId !== null ? 'demoPathwaysService' : 'pathwaysService')
```

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -44,7 +44,7 @@ const ExpandedNode: FC<ExpandedNodeProps> = ({
   const { patientRecords, setPatientRecords } = usePatientRecords();
   const client = useFHIRClient();
   const note = useNote();
-  const patient = usePatient();
+  const patient = usePatient().patient as fhir.Patient;
```
blangley28, line 47: The type of `usePatient().patient` is `fhir.Patient | null`. I think all this `as fhir.Patient` is doing is covering a null check. In my personal opinion we should make the null check explicit because it is easier to read but would like to hear from others too.

src/fixtures/DemoData.tsx
mdube7: file was removed

src/index.js
```diff
@@ -11,8 +11,12 @@ const rootElement = document.getElementById('root');
 ReactDOM.render(
   <Router>
     <Switch>
-      <Route exact path="/" render={() => <App demo={false} />}></Route>
-      <Route path="/demo" render={() => <App demo={true} />}></Route>
+      <Route exact path="/" render={() => <App demoData={null} />}></Route>
```
blangley28: Looks like `demoData` was removed from the props interface so it should be deleted here
```suggestion
      <Route exact path="/" render={() => <App  />}></Route>
```

blangley28:
The app doesn't seem to launch correctly from the smart launcher
Screen Shot 2020-04-06 at 10 02 19 AM

src/index.js
mdube7: updated to demoId={null}

src/components/App.tsx
mdube7: updated to !==

src/components/PatientProvider.tsx
danlee1025: This was changed to follow the same pattern as `PatientRecordsProvider`. The `patient` state was moved into the `App` component so that it would know about changes to the patient. It seems like no other component currently needs the `setPatient` function so we could remove it but in the future if other components need to change the patient we would need to add this function back.

mdube7:
> The app doesn't seem to launch correctly from the smart launcher

The smart launcher is working again now

src/components/ExpandedNode/ExpandedNode.tsx
danlee1025: I mean by the time we reach these components, `patient` will have already been set since the app will produce a loading screen instead of rendering the `children` components when `patient` is `null`.

src/index.js
blangley28: `demoId` is optional so it can be removed entirely and just be `<App />`

src/components/App.tsx
danlee1025: Agreed, we didn't always have Context for both and when I decided to create the `PatientRecords` one I didn't think to combine them.

blangley28 (approved):

dehall (approved):
Looks good, the only thing that seems a little weird is that if you start at http://localhost:3000/#/ and then go to http://localhost:3000/#/demo without refreshing, it uses the pathways from the service URL instead of from the demo location. I think that's just an artifact of how we're loading the pathway service URL and we can revisit later if that ever becomes an issue

PR #34, opened 2020-03-30
minimize renders

KeeyanGhoreshi:
Cuts down on some of the unnecessary rerenders.  For example every node would rerender twice on any action.  The graph would rerender unnecessarily when expanding nodes.

I made some components pure, which is not the ideal solution, but the gist is basically that the useEffect loop on the graph causes the graph & app to rerender a bunch.  The pure components avoid unnecessary rerenders by comparing props, so when the parent updates the child doesn't unless the props change.

One case I didn't fix is if you click on a node far along the pathway before you reach it, and fill in missing data, there's some wasted rerenders since the app updates the pathway object but doesn't change anything.  The app/graph rerender but nothing changes since the pathway doesn't update.


noranda (commented):
I'm confused why we're using `recompose` for this. Doesn't React's [`memo`](https://reactjs.org/docs/react-api.html#reactmemo) do the same thing? According to the [recompose](https://github.com/acdlite/recompose) github readme, the library is no longer being supported since it's become obsolete with the addition of hooks.

Re-renders happen when hooks get called within a component that renders something. We have a _lot_ of hooks. Pulling out those hooks into functional components separate from any component that renders anything, and wrapping the rendering components with `memo` should help. Then just pass in any context values it needs so it will only re-render if those values change.

KeeyanGhoreshi:
latest commit is the scenario noranda described, with the rendering component being wrapped with React.memo, and the parent component having all the hooks logic.  I only did graph because it has the most hooks and also I wanted to get some feedback if this is how we want to do our components for pretty much all of them.

src/components/Node/Node.tsx
```diff
@@ -33,86 +33,95 @@ const useStyles = makeStyles(
...
-          onClick={onClickHandler}
-        >
-          <div className={nodeStyles.iconAndLabel}>
-            <NodeIcon pathwayState={pathwayState} isGuidance={isGuidance} />
-            {label}
+      return (
+        <div className={topLevelClasses.join(' ')} style={style} ref={ref}>
+          <div
+            className={`${nodeStyles.nodeTitle} ${interactive && nodeStyles.clickable}`}
+            onClick={(): void => {
```
radamson: What was the reasoning behind replacing `onClickHandler` with its implementation?

Because the parent has to pass in both `interactive` and `setExpanded` anyways, and this is the only place it is used it would be better to keep it with the parent unless there is another reason.  That way the logic is only in 1 component instead of 2, and if we ever want to change the on click in other scenarios we only have to pass in a different handler instead of changing the Node code.

src/components/Graph/Graph.tsx
```diff
@@ -1,27 +1,18 @@
...
-import Arrow from 'components/Arrow';
 import { evaluatePatientOnPathway } from 'engine';
-import { EvaluatedPathway, PathwayResults, DocumentationResource } from 'pathways-model';
-import { Layout, NodeDimensions, Edge } from 'graph-model';
+import { EvaluatedPathway, PathwayResults } from 'pathways-model';
+import { Layout, NodeDimensions } from 'graph-model';
 import { usePatientRecords } from 'components/PatientRecordsProvider';
 import { DomainResource } from 'fhir-objects';
-
+import GraphRender from 'components/Graph/GraphRender';
```
radamson: use a relative import here because it's in the same directory.  GraphRender shouldn't be needed outside of Graph anyways since it's just an implementation detail.

```suggestion
import GraphRender from './GraphRender';
```

src/components/Graph/Graph.tsx
```diff
@@ -41,11 +32,12 @@ const Graph: FC<GraphProps> = ({
 
   const setPath = useCallback(
     (value: PathwayResults): void => {
-      _setPath(value.path);
+      if (!(value.path.toString() === path.toString())) _setPath(value.path);
```
radamson: I think using `!==` more clearly conveys what you're trying to do and is a little simpler.

```suggestion
      if (value.path.toString() !== path.toString()) _setPath(value.path);
```

Also why was this additional logic needed?

src/components/Graph/Graph.tsx
```diff
@@ -103,12 +95,14 @@ const Graph: FC<GraphProps> = ({
     });
   }
 
-  const initialExpandedState = useMemo(() => {
-    return Object.keys(layout).reduce((acc: { [key: string]: boolean }, curr: string) => {
+  const layoutKeys = Object.keys(layout);
+  const initialExpandedState = layoutKeys.reduce(
```
radamson: I'm ok with the change, but curious as to why you removed `useMemo` here.  Did it hurt performance or cause any issues?

src/components/Graph/Graph.tsx
```diff
@@ -174,70 +168,21 @@ const Graph: FC<GraphProps> = ({
...
+      documentation={documentation}
+      edges={edges}
+      path={path}
+      nodeRefs={nodeRefs}
+      windowWidth={windowWidth}
+      maxWidth={maxWidth}
+      pathway={pathway}
+      expanded={expanded}
+      setExpanded={setExpanded}
+    ></GraphRender>
```
radamson: Since there are no children you should use a self-closing tag
```suggestion
    />
```

src/components/Node/Node.tsx
radamson: Looking more it looks like doing this might allow the removal of the `_key` prop too.  but I might just be missing something 🙂 .

src/components/Node/Node.tsx
```diff
@@ -33,86 +33,95 @@ const useStyles = makeStyles(
...
-      top: yCoordinate,
-      left: xCoordinate
-    };
-
-    const backgroundColorClass = isOnPatientPath
-      ? styles.onPatientPath
-      : clsx(styles.notOnPatientPath, classes['not-on-patient-path']);
-    const isActionable = isCurrentNode && !documentation;
-    const topLevelClasses = [styles.node, backgroundColorClass];
+const Node: FC<NodeProps & { ref: Ref<HTMLDivElement> }> = React.memo(
```
radamson: since we typically import React functions explicitly (we don't use `React.useState`) I'd do that with memo too.

i.e. `import { memo} from 'react'`
and then just call `memo(...`

package.json
```diff
@@ -17,7 +17,9 @@
     "@types/react": "^16.9.17",
     "@types/react-dom": "^16.9.4",
     "@types/react-select": "^3.0.5",
+    "@types/recompose": "^0.30.7",
```
radamson: looks like `recompose` is not removed from the `package.json`.

radamson (commented):
I like the way you went about this by breaking out the parts of components that only depend on the props and memoizing them.  I did have a couple questions on some logic that looks like it changed that I was hoping you could help me understand.

noranda:
This looks good! Here's a [blog post](https://blog.axlight.com/posts/4-options-to-prevent-extra-rerenders-with-react-context/) that explains this approach a little better (Option 2). The Graph component could still use some cleanup not related to this task, but this works and should be done for our other components that also use shared contexts.

src/components/Node/Node.tsx
KeeyanGhoreshi: The onclickHandler causes the child to rerender.  The components passed through do not.  I realize now I could've memoized it but either way, this way prevents an unnecessary rerender of the child.

src/components/Graph/Graph.tsx
KeeyanGhoreshi: Because setPath doesn't just set the path, and it gets called while the value.path is the same but it still causes the component to rerender.  But we still might need the second part of setPath.

src/components/Graph/Graph.tsx
KeeyanGhoreshi: I'm honestly blanking on why I made this change.  I think it saves a render due to layout object changing but values being the same, causing useMemo to rerender the graph.

src/components/Node/Node.tsx
radamson: This sounds like a case, as I think you're alluding to, where we could use `useCallback` on `onClickHandler`. ([The React docs mention it is useful for passing callbacks to children while preventing unnecessary rerenders](https://reactjs.org/docs/hooks-reference.html#usecallback)).

I would suggest trying that approach.

src/components/Graph/Graph.tsx
radamson: Ah ok thanks that explains it!  I was thinking that the component wouldn't rerender if a `React.setState` call if the resulting state was the same, but after doing a little reading based on your comment it appears components unconditionally re-render after every setState call regardless.

Edit: As I read more it seems like react will not rerender children if the state is the same, but may render that specific component. They suggest using `useMemo` to avoid expensive calculations (which I can see how that would work when using the pure components like you're implementing here).
https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update

I don't mind the if statement here and I think it works well, but it seems like there are better, and more declarative, ways to avoid rerenders in the future than to always have to check the state before setting it once we've broken down the components a bit.

radamson: Any thoughts on the suggested code change?

src/components/Graph/Graph.tsx
radamson: Interesting.  I wonder if any reductions in rerenders are worth the loss of memoization (i.e. more rerenders that are faster overall, versus fewer rerenders that are slower overall).

Because Graph has so much state and does so much I wouldn't be surprised if the rerenders are more expensive.  It might make more sense to memoize if broken down a bit more, but that's a thought for another day.

src/components/Node/Node.tsx
KeeyanGhoreshi: I actually did use useCallback to begin with since that seemed like the most correct solution but it, too, caused a bonus render.

package.json
KeeyanGhoreshi: forgot about the types

src/components/Graph/Graph.tsx
KeeyanGhoreshi: the suggested change seems fine to me

src/components/Node/Node.tsx
radamson: hmm, ok.  I just want to make sure we're not losing some modularity just to optimize and I'm not sure how costly this render is.  Any tool we use is going to have some overhead, but its a tradeoff.

Seems smelly to be pushing down state into components that don't care about it and no longer being able to pass in an onClick Handler...

src/components/Graph/Graph.tsx
radamson: Also I wonder if we could be stepping on a landmine doing this because setState is called asynchronously (so the previous state we're comparing against might not be correct).

React provides a method of doing updates computed against the previous state for this using functional updates: https://reactjs.org/docs/hooks-reference.html#functional-updates

radamson: @KeeyanGhoreshi implemented a better solution here that moves the conditional check into a functional state update.

src/components/Graph/Graph.tsx
radamson: @KeeyanGhoreshi and I discussed and tested out some alternatives here together, but right now we can't reasonably memoize this without some greater refactoring because the dependency is only shallowly compared in React.

src/components/Node/Node.tsx
radamson: This is now working with useCallback.

radamson:
This makes some good improvements and I'm glad we were able to add back `onClickHandler` using `useCallback` to node.

Further performance updates may require greater decomposition of these components (especially large ones with a lot of state like `Graph` in order to really take advantage of the performance enhancing hooks like `useCallback` and `useMemo` and `memo`.  The biggest issue is that much of the data and state we are working with are complex objects and arrays. `useMemo`, `useCallback` and `memo` only do shallow comparisons.  Even `GraphRender` accepts a number of complex props as arguments which means it will still rerender more than we would like (this might not be completely avoidable, but we should keep this in mind as we design components).

Here's some good resources on this that I found as we were working through this:

https://twitter.com/dan_abramov/status/1104414272753487872
https://github.com/facebook/react/issues/17184
https://github.com/facebook/react/issues/14476#issuecomment-471199055
https://reactjs.org/docs/react-api.html#reactmemo

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -327,4 +285,69 @@ function renderGuidance(
   return returnElements;
 }
 
+interface RenderProps {
```
radamson: Sorry to add more comments, but we should rename this to match the style guide: https://github.com/mcode/pathways/wiki/Code-Style-Guide#props

```suggestion
interface ExpandedNodeRenderProps {
```

src/components/Graph/Graph.tsx
radamson: @KeeyanGhoreshi after the rebase I think this updated code was lost.  I don't see the original if statement added or the functional state update...

src/components/Graph/Graph.tsx
```diff
@@ -30,240 +45,298 @@ const isEdgeOnPatientPath = (pathwayResults: PathwayResults, edge: Edge): boolea
...
+  parentWidth: number;
+  maxWidth: number;
+  pathway: Pathway;
+  expanded: {
+    [key: string]: boolean | undefined;
+  };
+  setExpanded: (key: string, expand?: boolean | undefined) => void;
+}
+
+const GraphRender: FC<GraphRenderProps> = memo(
```
radamson: You probably hate me by now, but I think the suffix "Memo" would more accurately describe the component.  i.e. `GraphMemo`

radamson (approved):
Nice job @KeeyanGhoreshi and @kghoreshi.

dehall:
I'm good with the changes here, but originally I think we were hoping this would cut down on some of the jumpiness in the UI, like when originally opening a pathway or when accepting an action. Based on our discussion (last week?) it seems like that's not going to happen, but @KeeyanGhoreshi it sounded like you had some ideas of what we might have to do instead. Could you add a backlog item for that and add any thoughts you have on how we'd make that happen?

KeeyanGhoreshi:
I thought we already had a backlog task for that, with batch updates to reduce jumpiness.

dehall (approved):

PR #35, opened 2020-03-30
Patient Record Viewer

blangley28:
Update the patient record viewer to match the new designs. 

Updated so the pathway stays in the center of the container even when window is resized or the container size changes (such as opening the record viewer). This will come in handy too for resizing the graph in the pathway list view

Screen Shot 2020-03-30 at 3 34 17 PM
Screen Shot 2020-03-30 at 3 34 35 PM
Screen Shot 2020-03-30 at 3 35 00 PM

### DIFF
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -197,7 +197,6 @@ const Graph: FC<GraphProps> = ({
         height: interactive ? maxHeight + 150 : 'inherit',
         width: maxWidth + 'px',
         position: 'relative',
-        overflow: 'auto',
         marginRight: '5px'
       }}
     >
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -44,9 +44,6 @@ const Graph: FC<GraphProps> = ({
   const [parentWidth, setParentWidth] = useState<number>(
     graphElement?.current?.parentElement?.clientWidth ?? 0
   );
-  const [path, _setPath] = useState<string[]>(
-    evaluatedPathway.pathwayResults ? evaluatedPathway.pathwayResults.path : []
-  );
 
   const setPath = useCallback(
     (value: PathwayResults): void => {
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -8,6 +8,7 @@ import { EvaluatedPathway, PathwayResults, DocumentationResource } from 'pathway
 import { Layout, NodeDimensions, Edge } from 'graph-model';
 import { usePatientRecords } from 'components/PatientRecordsProvider';
 import { DomainResource } from 'fhir-objects';
+import styles from './Graph.module.scss';
 
 interface GraphProps {
   evaluatedPathway: EvaluatedPathway;
@@ -257,6 +258,7 @@ const Graph: FC<GraphProps> = ({
 };
 
 function useWindowWidth(): number {
+  // TODO: This needs to calculate the width of the parent element ('graph-root'), not the browser
   const getWidth = (): number => window.innerWidth;
   const [windowWidth, setWindowWidth] = useState(getWidth);
 
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -166,6 +166,16 @@ const Graph: FC<GraphProps> = ({
     setLayout(getGraphLayout());
   }, [expanded, getGraphLayout]);
 
+  // Set the height of the graph container
+  useEffect(() => {
+    const graphContainer = document.getElementById('graphContainer');
+    const headerHeight = document.getElementById('header')?.clientHeight;
+    const navHeight = document.getElementById('navigation')?.clientHeight;
+    console.log(graphContainer);
+    if (graphContainer && navHeight && headerHeight)
+      graphContainer.style.height = window.innerHeight - (navHeight + headerHeight) + 'px';
+  }, []);
+
   // maxWidth finds the edge label that is farthest to the right
   const maxWidth: number =
     edges !== undefined
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -14,6 +14,7 @@ interface GraphProps {
   evaluatedPathway: EvaluatedPathway;
   interactive?: boolean;
   expandCurrentNode?: boolean;
+  headerElement?: React.RefObject<HTMLDivElement>;
   updateEvaluatedPathways: (value: EvaluatedPathway) => void;
 }
 
@@ -33,6 +34,7 @@ const Graph: FC<GraphProps> = ({
   evaluatedPathway,
   interactive = true,
   expandCurrentNode = true,
+  headerElement = null,
   updateEvaluatedPathways
 }) => {
   const patientRecords = usePatientRecords();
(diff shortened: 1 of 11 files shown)


src/components/Graph/Graph.tsx
```diff
@@ -160,6 +161,15 @@ const Graph: FC<GraphProps> = ({
     setLayout(getGraphLayout());
   }, [expanded, getGraphLayout]);
 
+  // Set the height of the graph container
+  useEffect(() => {
+    const graphContainer = document.getElementById('graphContainer');
```
KeeyanGhoreshi: this feels like it could be a `ref` and avoid reaching into the dom directly.

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -14,29 +14,120 @@ import {
...
+
+  const expand = (): void => {
+    setIsExpanded(!isExpanded);
+  };
+
+  // Set the height of the patient record container
+  useEffect(() => {
+    const recordContainer = document.getElementById('recordContainer');
+    const headerHeight = document.getElementById('header')?.clientHeight;
+    const navHeight = document.getElementById('navigation')?.clientHeight;
```
KeeyanGhoreshi: Same thing here, are we able to pass refs to these elements through the components?  Feels like that would be the most React-appropriate way to go about doing this.  Though I personally don't see anything wrong with accessing elements like this, I do know its slower and causes some rendering inefficiencies from when I was implementing sticky in IE.  Not really noticeable on chrome but it could cause half a frame of lag on a samsung smart fridge running IE7, for example.

src/components/Graph/Graph.tsx
blangley28: Updated to use refs

src/components/PatientRecord/PatientRecord.tsx
blangley28: Updated to using refs. Do you have a samsung smart fridge running IE7 I can test on?

KeeyanGhoreshi: I don't commune with the smart-fridgeless, sorry

KeeyanGhoreshi:
A couple general comments:

It seems like there's a couple small things that don't align with the design, namely the pagination of the lists of resources over scrollbars, the number of resources being shown, and the headers.

I think the "observations" header should be removed since it's redundant:
image

src/components/PatientRecord/PatientRecord.module.scss
```diff
@@ -0,0 +1,76 @@
+@import '../../styles/variables';
+
+// ----------  TITLE SPACING  --------- //
+$verticalPadding: 20px;
+$lineHeight: 30px;
+
+.record {
```
KeeyanGhoreshi, line 7: This might be something that has to be addressed in the fhir-visualiziers, but I'll still mention it here, LOINC codes are being broken up due to the dash, since the text wraps:

image

We could avoid this by using a non-breaking dash character, or by giving the first column some more width.  As it stands the dash counts as a line-breaking character and the column is only as wide as the word "LOINC".

blangley28: The resource tabs are just using the fhir  visualizers as a place holder right now. Their designs are going to change in a future task so I don’t see it as a huge deal

KeeyanGhoreshi: makes sense

blangley28:
All of the data in the sidebar is just from the existing fhir visualizers. The structure is there so we can create other panels as needed when the data is available in the system. 

As for the header - this is once again from fhir visualizers which is just a place holder. I'm hesitant to remove this from the visualizers because if they are used in another context or repo then the header would be missing and there would be no indication to the resource type.

src/components/PatientRecord/PatientRecord.module.scss
```diff
@@ -0,0 +1,76 @@
...
+        padding: 25px;
+        background-color: $gray;
+        overflow-y: scroll;
+    }
+
+    .element {
+        display: block;
+        margin-bottom: 1vh;
+        font-size: 1.1em;
+        text-transform: uppercase;
```
dehall: This seems to be transforming all text, not just the headers. Per the design I think we want the internal content to retain its original case.
image

Also this is very subjective but the all-caps section headers look a little odd to me.
image


 I can't quite put my finger on what specifically the difference between the design and implementation is but somehow the design looks "friendlier"

blangley28: Updated text-transform to be on the title only. Increased the font size and added a small letter spacing to make it look friendlier
Screen Shot 2020-04-01 at 8 57 31 AM

blangley28:
Title was missing the element count. Added the number of resources to the title for all except patient
Screen Shot 2020-04-01 at 9 08 20 AM

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -14,29 +14,123 @@ import {
...
 const getResourceByType = (
   patientRecord: ReadonlyArray<DomainResource>,
   resourceType: string
 ): ReadonlyArray<object> => {
   return patientRecord.filter(resource => resource.resourceType === resourceType);
 };
 
-const PatientRecord: FC = () => {
+interface PatientRecordProps {
+  headerElement: React.RefObject<HTMLDivElement>;
```
radamson: would be nice to import this the same way for other React bindings.
i.e. on line 1
`import React, { FC, useState, useEffect, useRef, RefObject } from 'react';`

And then here:
```suggestion
  headerElement: RefObject<HTMLDivElement>;
```

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -14,29 +14,123 @@ import {
...
   const patient = usePatient();
   const resources = usePatientRecords().patientRecords;
+  const resourcesByType = getResourceByType(resources, resourceType);
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+
+  const chevron: IconProp = isExpanded ? faChevronUp : faChevronDown;
+  const resourceCount: string =
+    resourceType !== 'Patient' ? '(' + resourcesByType.length + ')' : '';
+
+  const visualizer = (): JSX.Element | undefined => {
```
radamson: we typically use `ReactNode` as a return type for these types of things which includes `undefined`.
```suggestion
  const visualizer = (): ReactNode => {
```

Also why not break this out into another component (that would set it up to be memoized it for a performance boost too!)?  as it stands this function will be redefined on every rerender.  Breaking to code up is almost always beneficial 🙂

src/components/PatientRecord/PatientRecord.tsx
blangley28: Updated import

src/components/PatientRecord/PatientRecord.tsx
blangley28: Good call - created new FC Visualizer

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -14,31 +14,133 @@ import {
...
+          <Visualizer resourceType={resourceType} resourcesByType={resourcesByType} />
+        </div>
+      )}
     </div>
   );
 };
 
+const Visualizer: FC<VisualizerProps> = ({ resourceType, resourcesByType }) => {
+  const patient = usePatient();
+  if (resourceType === 'Patient') return <PatientVisualizer patient={patient} />;
```
dehall, line 130: I won't hold up the PR for just this, but anytime I see a long chain of if-else-if on a single variable I think it should be a `switch` instead. Since everything here is so similar it might also be possible to do a "dynamic type", though not sure if that's possible or easy with TypeScript:
```       
const Something = ...;
return <Something rows={resourcesByType} />
```

dehall (approved):

KeeyanGhoreshi (approved):

radamson (approved):
Nice!

Give it one last rebase and I'll merge it in.

blangley28:
Rebased

PR #36, opened 2020-03-31
Note Preview

radamson:
Implements the note preview.  The main thing missing now is the information in the clinical section which is not readily available in the note as far as I know.  Another thing we may want to consider related to this is hiding any fields which are not populated

Screen Shot 2020-03-31 at 8 47 24 AM


public/config.js
```diff
@@ -1,5 +1,5 @@
 CONFIG = {
-  mapper: 'SyntheaToV09', // mapper name from fhir-mapper -- allows mapping of incoming FHIR resources
+  mapper: null, // mapper name from fhir-mapper -- allows mapping of incoming FHIR resources
```
blangley28: Do we want to push this change @dehall?

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -25,6 +26,7 @@ import {
   MedicationRequest,
   ServiceRequest
 } from 'fhir-objects';
+import ReportModal from '../ReportModal';
```
blangley28: Use absolute import
```suggestion
import ReportModal from 'components/ReportModal';
```

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -53,7 +60,13 @@ const ExpandedNode: FC<ExpandedNodeProps> = ({
 
     // Create DocumentReference and add to patient record(and post to FHIR server)
     if (note) {
-      const noteString = createNoteContent(note, patientRecords, status, comments, pathwayState);
+      const noteString = createNoteContent(
+        note,
+        patientRecords,
+        status,
+        comments ?? '',
```
blangley28: Is this needed? It seems like we can just grab the comments from `note.notes`

public/config.js
dehall: Yeah we do, there's no real benefit to using the mapper at the moment, and potential downside in that it can cause exceptions. I also changed it to null in my current PR.

src/components/ReportModal/ReportSection/ReportSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const ReportSection: FC<ReportSectionProps> = ({ icon, fields }) => {
+  return (
+    <div className={styles.sectionContainer}>
+      <div className={styles.sectionIconContainer}>
+        <FontAwesomeIcon icon={icon} className={styles.icon} />
+      </div>
+      <div className={styles.fieldContainer}>
+        {fields.map(field => (
+          <ReportField sectionTitle={field.name} value={field.value} />
```
blangley28: This should probably include a `key`.
```suggestion
          <ReportField sectionTitle={field.name} value={field.value} key={field.name} />
```

src/components/ReportModal/ReportModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+import ActionButton from 'components/ActionButton';
+
+interface ReportModalInterface {
+  onConfirm: () => void;
+  onDecline: () => void;
+}
+
+const ReportModal: FC<ReportModalInterface> = ({ onConfirm, onDecline }) => {
+  const note = useNote();
+  const reportName = `${note?.pathway} Report`;
```
blangley28: Up for discussion since I know the design says "{Pathway name} Report" but I think this should be "{State name} Report" since we are submitting a note to accept a specific state

blangley28:
First off looks really great! Two notes about functionality:
1. Do we want this to pop up on decline as well? It makes sense we would to verify the note etc and show the status as declined.
2. Under the pathway section the recommendation always seems to be empty. I'm not sure when the `note.node` field is set but I worry it is after you accept - which means the next time we go to add a note it will have the previous node in it. I think the status in this section should also say accepted/declined not "pending"

src/components/ExpandedNode/ExpandedNode.tsx
radamson: Can't remember why I did this, but it doesn't seem necessary.  Changed, let me know what you think.

src/components/ReportModal/ReportSection/ReportSection.tsx
radamson: Good call I missed that error somehow.

It was a little trickier than just using `field.name` since it is typed as a `ReactNode` and `key` requires a `string | number | undefined`.  I think it's desirable to leave name as a `ReactNode` so that a complex HTML element can be used as a name (maybe something in bold or with an icon).

instead of using the index as a key ([generally bad](https://reactjs.org/docs/lists-and-keys.html#keys)) or manually specifying the keys I opted to add and use `shortid` which is the preferred option in the article linked in the react docs.

I also did add filtering for any section that has a falsey name while working on this.

src/components/ReportModal/ReportModal.tsx
radamson: That makes more sense to me.  @dehall Thoughts? ☕️

radamson:
Good points.  Right now it looks like the note context doesn't ever really get updated with all the required information (like the note.node).  Instead this information is just put into the note string right before it is written back

https://github.com/mcode/pathways/blob/107bd60d4734da7d1bd14073cb50a6eb3f2d2598/src/components/ExpandedNode/ExpandedNode.tsx#L51-L60

IMO it would make more sense to keep the note update to date with the pathway, comments, and pathwayStatus and `onConfirm` will only be responsible for creating and writing the DocumentReference.  `createNoteContent` seems to be doing some of the job that the note context should really be handling.

Otherwise I could pass these missing things into the ReportModal as props, but it sorta defeats the purpose of the note context if it contains some of the note data, but not all (including some of the fields it has already declared!).

src/components/ReportModal/ReportModal.tsx
dehall: TBH both feel a little awkward to me though I won't strongly oppose either. Not to bikeshed but personally I would lean toward just "Pathway Report". The pathway and state name should both be included here anyway right?

dehall:
So I agree with Ben, we probably want this same preview to show up for Decline as well, and then in this section:
image
Recommendation should be the name of the current state, and status should either be Accepted or Declined

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -88,7 +95,7 @@ const ExpandedNode: FC<ExpandedNodeProps> = ({
             className={`${indexStyles.button} ${styles.defaultTextButton}`}
             onClick={(e): void => {
               e.preventDefault();
-              if (!comments.includes(defaultText)) setComments(comments + defaultText);
+              if (!comments?.includes(defaultText)) setComments(comments + defaultText);
```
blangley28: Since comments is initialized to be `note?.notes ?? ''` it will always be nonnull
```suggestion
              if (!comments.includes(defaultText)) setComments(comments + defaultText);
```

src/components/ReportModal/ReportModal.tsx
radamson: Changed the name to "Pathway Report"

radamson:
I'm thinking about moving the NoteDataProvider down so that each node has its own context (right now we only have one that wraps the whole app).  The notes themselves, including the comments and whether they are accepted/declined are really tied to the individual nodes though.

src/components/ExpandedNode/ExpandedNode.tsx
blangley28: Did you remove the `?` and then add it back in? `comments` is still initialized as `note?.notes ?? ''` so I still don't think it is needed.

radamson: No, I think I just hosed that on the rebase. Fixed.

blangley28:
Looks like you addressed all of my comments. Two more bugs I found though:

1. It looks like the actions are always accepted (even when clicking the decline button and the node saying decline). I tested in the latest master and this is not happening there.
2. The note `z-index` should be increased to go ontop of all nodes. I expanded a node by accident and then opened the note and saw this:
Screen Shot 2020-04-09 at 9 26 38 AM
My guess is we increase the z-index of manually expanded nodes which causes it to render on top of the note preview

radamson:
Nice catches @blangley28.  Working on fixes now...

dehall:
I think the point about moving the note data provider down is a good one. In the "multiple options" scenario if you type in one note field then accept a different action, the notes from the first action show up. (Screenshots below. It's not a realistic user scenario but it's a sign that things are linked in ways we don't intend for them to be)
That said I don't know if it makes sense for it to be changed here since that relationship already exists. If it's really easy to change though go ahead, otherwise let's make it a backlog item.


image

-->

image

src/components/ReportModal/ReportModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const reportName = 'Pathway Report';
+  const patientSection = [
+    { name: 'Date', value: note?.date },
+    { name: 'Patient', value: note?.patient },
+    { name: 'Date of Birth', value: note?.birthdate },
+    { name: 'Physician', value: note?.physician }
+  ];
+
+  const observationField = [
+    { name: 'Primary Cancer:', value: undefined },
```
dehall: Super minor but the colon after the name here makes it show up twice in the actual modal

radamson:
@blangley28 issues are addressed.  Thanks for catching those!

dehall:
The latest changes did something to the note where it's keeping the text of the previous action. For example on the demo patient, neoadjuvant pathway, choose T > 2, then the three options show up with "T > 2" in the note field:
image

radamson:
> The latest changes did something to the note where it's keeping the text of the previous action. For example on the demo patient, neoadjuvant pathway, choose T > 2, then the three options show up with "T > 2" in the note field:
> image

@dehall previously the comments were not synchronized at initialization with the notes in the NoteContext (if there were notes already in the note context they wouldn't be seen and just ovewritten when the node was accepted).  Now the comments are synchronized with the note.

The MissingDataPopup actually sets the notes to the selected value: https://github.com/mcode/pathways/blob/e2628d65bbd4910161abb22bd46b299f09eff42d/src/components/MissingDataPopup/MissingDataPopup.tsx#L83

https://github.com/mcode/pathways/blob/e2628d65bbd4910161abb22bd46b299f09eff42d/src/utils/fhirUtils.ts#L107-L113

So now that the comments are synchronized and because the NoteContext closes over all nodes we see it.

Thanks for pointing this out, I actually noticed this too and am working on fixing it by moving the NoteContext.

It might be worth holding off on reviewing for a bit @blangley28 and @dehall because I think it needs more updates than I expected...

radamson:
@dehall moved the `NoteDataProvider` so it individually wraps Nodes for unique notes so that issue should be fixed.

dehall (approved):
Looks great. Can you add a backlog item to populate the missing fields in the note? Maybe we want like an "mCODE Context" or something so we can use it here in the note and on the patient record sidebar

radamson:
@dehall created backlog item 223 to populate the missing fields.  Also I added another commit this morning that removes the `NoteDataProvider` from `App` and adds a `UserProvider` for sharing the physician information to the notes.

src/components/NoteDataProvider.tsx
```diff
@@ -23,25 +32,36 @@ interface NoteDataProviderProps {
...
+}
+
+export const NoteContext = createContext<NoteContextProps>({
+  note: null,
+  setNote: () => {
+    // do nothing.
+  }
+});
 
 export const NoteDataProvider: FC<NoteDataProviderProps> = ({ children, date, physician }) => {
```
dehall: Now that the note below uses the UserProvider, can the physician prob be removed?

radamson: Good call 👍

blangley28 (approved):

PR #37, opened 2020-04-02
Bump fhir-visualizers version

danlee1025:
- @dehall re-deployed an updated version of fhir-visualizers to npm and I bumped the version to 0.0.2 in `package.json`.

The error is actually no longer occurring on `master` since the neoadjuvant pathway now creates a proper `MedicationRequest`.  I added the old version onto the pathways server so that this can be tested.

- Launch the app via [SMART](http://moonshot-dev.mitre.org:4001/?auth_error=&fhir_version_1=r2&fhir_version_2=r4&iss=&launch_ehr=1&launch_url=http%3A%2F%2Flocalhost%3A3000%2F&patient=&prov_skip_auth=1&provider=&pt_skip_auth=1&public_key=&sb=&sde=&sim_ehr=0&token_lifetime=15&user_pt=)
- Choose a patient and choose the old neoadjuvant pathway
- Accept any chemotherapy regimen
- Open the patient viewer and view the list of medicationrequests

Should see something like this
image
The app would previously crash when it couldn't get the code and display off of the medication resource. This version of fhir-visualizers catches the error and prevents the crash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "dagre": "^0.8.5",
     "eslint-plugin-only-error": "^1.0.2",
     "fhir-mapper": "^3.1.1",
-    "fhir-visualizers": "^0.0.1",
+    "fhir-visualizers": "~0.0.2",
     "fhirclient": "^2.1.0",
     "node-sass": "^4.12.0",
     "react": "^16.10.1",
--- a/src/engine/output-results.ts
+++ b/src/engine/output-results.ts
@@ -169,7 +169,7 @@ function getConditionalNextState(
   for (const transition of currentState.transitions) {
     if (transition.condition) {
       let currentTransitionDocumentation: DocumentationResource | null = null;
-      if (patientData[transition.condition.description].length)
+      if (patientData[transition.condition.description]?.length)
         // TODO: add functionality for multiple resources
         currentTransitionDocumentation = patientData[transition.condition.description][0];
       else {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "dagre": "^0.8.5",
     "eslint-plugin-only-error": "^1.0.2",
     "fhir-mapper": "^3.1.1",
-    "fhir-visualizers": "^0.0.1",
+    "fhir-visualizers": "~0.0.2",
```
blangley28, line 28: Maybe my understanding of this syntax is incorrect but shouldn't `^` have installed 0.0.2 since it is a patch?

radamson: I think you're correct @blangley28 from [looking at the docs](https://github.com/npm/node-semver#tilde-ranges-123-12-1).  Not sure one is better than the other, but I suppose `~` is a little more conservative.

dehall: I seem to recall there's something special about the versioning for initial patch releases. 0.0.1 -> 0.0.2 is treated differently than 1.1.1 -> 1.1.2 for example. Not sure about the difference between `^` and `~` but either way I think the manual bump to 0.0.2 is necessary.

danlee1025: I see, that's interesting... Did we want `^0.0.2` or `^~0.0.2`?

blangley28 (approved):

package.json
blangley28: Is `^~0.0.2` valid syntax? I think `^0.0.2` should do it

dehall (approved):

PR #33, opened 2020-03-25
Updated demo patient to support happy path scenario

danlee1025:
- Added `DocumentReference` to simulate selecting missing data in her2 pathway. Tumor size node is completed.

image

PR #38, opened 2020-04-07
Delete old patient fixture and update new file with prior fix

dehall:
Deletes the old patient fixture file which was no longer used given the demo URL switch, and reintroduces the fix from #33


mdube7 (approved):
Looks good to me

danlee1025 (approved):
Looks good, I see both scenarios on the Maureen demo endpoint.

PR #39, opened 2020-04-07
Node documentation

blangley28:
Loads documentation for nodes not on the resource (both documentation references and actual documentation). Currently marks with a check if complete to indicate to the oncologist it is pending.
Screen Shot 2020-04-07 at 4 22 50 PM

### DIFF
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -33,14 +33,24 @@ interface GraphProps {
   updateEvaluatedPathways: (value: EvaluatedPathway) => void;
 }
 
+const getPath = (pathwayResults: PathwayResults): string[] => {
+  return Object.entries(pathwayResults.documentation)
+    .filter(entry => {
+      const [, doc] = entry;
+      return doc.onPath === true;
+    })
+    .map(entry => {
+      const [, doc] = entry;
+      return doc.state;
+    });
+};
+
 const isEdgeOnPatientPath = (pathwayResults: PathwayResults, edge: Edge): boolean => {
-  const startIndex = pathwayResults.path.indexOf(edge.start);
-  const endIndex = pathwayResults.path.indexOf(edge.end);
+  const path = getPath(pathwayResults);
+  const startIndex = path.indexOf(edge.start);
+  const endIndex = path.indexOf(edge.end);
   if (startIndex !== -1 && endIndex !== -1 && startIndex + 1 === endIndex) return true;
-  else if (
-    startIndex === pathwayResults.path.length - 1 &&
-    pathwayResults.currentStates.includes(edge.end)
-  )
+  else if (startIndex === path.length - 1 && pathwayResults.currentStates.includes(edge.end))
     return true;
   else return false;
 };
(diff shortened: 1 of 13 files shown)


src/components/Graph/Graph.tsx
```diff
@@ -18,14 +18,24 @@ interface GraphProps {
...
+const getPath = (pathwayResults: PathwayResults): string[] => {
+  return Object.entries(pathwayResults.documentation)
+    .filter(entry => {
+      const [, doc] = entry;
+      return doc.onPath === true;
+    })
+    .map(entry => {
+      const [, doc] = entry;
+      return doc.state;
+    });
```
dehall: Since you're never using the keys from these entries, and `doc.onPath` is guaranteed to be a boolean already, can this be simplified to:
```suggestion
  return Object.values(pathwayResults.documentation)
    .filter(doc => doc.onPath)
    .map(doc => doc.state);
```

src/engine/output-results.ts
```diff
@@ -119,6 +124,81 @@ export function criteriaData(pathway: Pathway, patientData: PatientData): Criter
...
+ * @param patientDocumentation - the documentation dictionary from states on the patient path
+ */
+function getAllDocumentation(
+  pathway: Pathway,
+  patientData: PatientData,
+  resources: DomainResource[],
+  patientDocumentation: { [key: string]: Documentation }
+): { [key: string]: Documentation } {
+  const statesWithDocumentation = Object.entries(patientDocumentation)
+    .filter(entry => {
```
dehall: Same simplification as above should be possible here

dehall (approved):
Overall looks good. I would have been fine with leaving the checkmark icon as-is, and not displaying them on the branch states, but this works. A couple minor refactor suggestions

mdube7:
This might be behaving as anticipated but I wanted to call attention to have it verified
In the Maryam demo the `node status prior to surgery` says missing data. 
If you select T<=2 and then N+ or N0 the data will show up in `node status prior to surgery`.
If you select T>2 I can't seem to get the missing data in `node status prior to surgery` to appear.
image

blangley28:
> This might be behaving as anticipated but I wanted to call attention to have it verified
> In the Maryam demo the `node status prior to surgery` says missing data.
> If you select T<=2 and then N+ or N0 the data will show up in `node status prior to surgery`.
> If you select T>2 I can't seem to get the missing data in `node status prior to surgery` to appear.

So that is expected. The "Node Status Prior to Surgery" node is the same data as the "Node Status" node. Therefore, selecting N+/N0 in the "Node Status" node _should_ populate the data for the "Node Status Prior to Surgery" node. When selecting T > 2 on "Tumor Size" the user does not input the data under "Node Status" so nothing will appear under "Node Status Prior to Surgery"

dehall:
I'm actually surprised that selecting the missing data for "Node Status" affects the later state "Node Status Prior to Surgery" ... I thought the note written for missing data was based on state name?

mdube7:
looks good to me

blangley28:
I believe it is based on the CQL which for action states is the state name and for branch states is the branch condition

src/components/Graph/Graph.tsx
blangley28: Updated here and in output-results

mdube7 (approved):

PR #40, opened 2020-04-09
Add pathway criteria to patient record

blangley28:
Added "Pathway" to the top of the patient record with the pathway criteria. Note this does require calling the Pathway Criteria again but [PATHWAYS-203](https://jira.mitre.org/browse/PATHWAYS-203) already exists in the backlog to address this concern.
Screen Shot 2020-04-09 at 1 41 09 PM

### DIFF
--- a/src/components/PatientRecord/PatientRecord.tsx
+++ b/src/components/PatientRecord/PatientRecord.tsx
@@ -27,11 +27,40 @@ import { CriteriaResult } from 'pathways-model';
 import { usePathwayContext } from 'components/PathwayProvider';
 import { evaluatePathwayCriteria } from 'engine';
 
-const getResourceByType = (
-  patientRecord: ReadonlyArray<DomainResource>,
-  resourceType: string
-): ReadonlyArray<object> => {
-  return patientRecord.filter(resource => resource.resourceType === resourceType);
+const resourceTypes = [
+  'Pathway',
+  'Patient',
+  'Condition',
+  'Observation',
+  'DiagnosticReport',
+  'MedicationRequest',
+  'AllergyIntolerance',
+  'CarePlan',
+  'Procedure',
+  'Encounter',
+  'Immunization'
+];
+
+const groupResourceByType = (
+  patientRecord: ReadonlyArray<DomainResource>
+): ReadonlyMap<string, ReadonlyArray<DomainResource>> => {
+  const map: Map<string, DomainResource[]> = new Map();
+  patientRecord.forEach(resource => {
+    const resourceType = resource.resourceType ?? '';
+    if (resourceTypes.includes(resourceType)) {
+      const collection = map.get(resourceType);
+      if (!collection) map.set(resourceType, [resource]);
+      else collection.push(resource);
+    }
+  });
+  return map;
+};
+
+const getResourcesByType = (
+  resourceType: string,
+  groupedResources: ReadonlyMap<string, ReadonlyArray<DomainResource>>
+): ReadonlyArray<DomainResource> => {
+  return groupedResources.get(resourceType) ?? [];
 };
 
 interface PatientRecordProps {
@@ -40,29 +69,19 @@ interface PatientRecordProps {
 
 interface PatientRecordElementProps {
   resourceType: string;
+  resources: ReadonlyArray<DomainResource>;
 }
 
 interface VisualizerProps {
   resourceType: string;
-  resourcesByType: readonly object[];
+  resourcesByType: ReadonlyArray<DomainResource>;
 }
 
 const PatientRecord: FC<PatientRecordProps> = ({ headerElement }) => {
+  const resources = usePatientRecords().patientRecords;
   const recordContainerElement = useRef<HTMLDivElement>(null);
   const [isExpanded, setIsExpanded] = useState<boolean>(false);
-  const resourceTypes = [
-    'Pathway',
-    'Patient',
-    'Condition',
-    'Observation',
-    'DiagnosticReport',
-    'MedicationRequest',
-    'AllergyIntolerance',
-    'CarePlan',
-    'Procedure',
-    'Encounter',
-    'Immunization'
-  ];
+  const groupedResources = groupResourceByType(resources);
 
   const expand = (): void => {
     setIsExpanded(!isExpanded);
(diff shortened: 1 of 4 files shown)


src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -127,7 +132,9 @@ const PatientRecordElement: FC<PatientRecordElementProps> = ({ resourceType }) =
 
 const Visualizer: FC<VisualizerProps> = ({ resourceType, resourcesByType }) => {
   const patient = usePatient().patient as fhir.Patient;
-  if (resourceType === 'Patient') return <PatientVisualizer patient={patient} />;
+
+  if (resourceType === 'Pathway') return <PathwayVisualizer />;
```
dehall, line 157: So the first time I saw this I thought a `switch` would be appropriate, but now I have another thought. We're creating a list of resource types above and then iterating over it and using this big if to pick elements out later. How would you feel about instead having it laid out like this in the PatientRecord:

```
<PatientRecordElement resourceType="Patient">
  <PatientVisualizer patient={patient} />
</PatientRecordElement>
<PatientRecordElement resourceType="Condition">
  <ConditionsVisualizer rows={getResourcesByType("Condition"} />
</PatientRecordElement>
...
```
(with variables re-oriented as needed)
In other words, the PatientRecordElement wouldn't know which visualizer to render, it would just be passed in as a child element to be rendered.

(I grant that ultimately it doesn't change much, I just don't like this giant list of ifs 😛 )

blangley28: One of the downsides of this is that `PatientRecordElement` already calls `getResourcesByType` so this would do the search twice. It already bothers me this is r*O(n) where r is the number of resource types and n is the number of resources in the patient record (ideally we would make it O(n)). 
I can't think of a _clean_ (or at least cleaner than the if block) way to share the result of `getResourcesByType` in this structure.

dehall: We can use a group-by function to make it O(n), lodash has one and I think we include lodash indirectly as a dependency of something else, or see https://stackoverflow.com/questions/14446511/most-efficient-method-to-groupby-on-an-array-of-objects

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -143,4 +150,39 @@ const Visualizer: FC<VisualizerProps> = ({ resourceType, resourcesByType }) => {
   else return <div>Unsupported Resource</div>;
 };
 
+const PathwayVisualizer: FC = () => {
+  const resources = usePatientRecords().patientRecords;
+  const evaluatedPathway = usePathwayContext().evaluatedPathway;
+  const [criteria, setCriteria] = useState<CriteriaResult | null>(null);
+
+  useEffect(() => {
+    // Create a fake Bundle for the CQL engine and check if patientPath needs to be evaluated
```
radamson: the word `fake` in here confused me for a sec as I was thinking it didn't actually have the patient records in it, but it looks like it does (unless I'm just reading it wrong).
```suggestion
    // Create a Bundle for the CQL engine and check if patientPath needs to be evaluated
```

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -143,4 +150,39 @@ const Visualizer: FC<VisualizerProps> = ({ resourceType, resourcesByType }) => {
...
+    const patient = {
+      resourceType: 'Bundle',
+      type: 'searchset',
+      entry: resources.map((r: fhir.Resource) => ({ resource: r }))
+    };
+
+    // Evaluate pathway criteria
+    if (evaluatedPathway) {
+      evaluatePathwayCriteria(patient, evaluatedPathway.pathway).then(criteriaResult =>
```
radamson, line 189: We might want to add a ticket for creating a `CriteriaContext` to avoid recalculating the criteria multiple times in the future.  `PathwayList` also calculates it currently.  Doesn't have to be done in this PR though.

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -102,8 +106,9 @@ const PatientRecordElement: FC<PatientRecordElementProps> = ({ resourceType }) =
   const [isExpanded, setIsExpanded] = useState<boolean>(false);
 
   const chevron: IconProp = isExpanded ? faChevronUp : faChevronDown;
-  const resourceCount: string =
-    resourceType !== 'Patient' ? '(' + resourcesByType.length + ')' : '';
+  const resourceCount: string = !['Patient', 'Pathway'].includes(resourceType)
+    ? '(' + resourcesByType.length + ')'
```
radamson: prefer template literal
```suggestion
    ? `(${resourcesByType.length})`
```

radamson (commented):
Looks good @blangley28 👍

src/components/PatientRecord/PatientRecord.tsx
dehall: Yeah I think that fits as part of -203 in the backlog, I'll bump that up

src/components/PatientRecord/PatientRecord.tsx
blangley28: Updated. This comment was copied so it was updated in `Graph.tsx` and `PathwayList.tsx` as well

src/components/PatientRecord/PatientRecord.tsx
blangley28: So I implemented the long list and the group by function but honestly I think it is 10 times cleaner using the big if statement. The more I write the more I'm itching to change it back because it looks harder to read IMO
```
<PatientRecordElement
       resourceType={'Condition'}
       resourceCount={getResourceByType('Condition', groupedResources).length}
>
            <ConditionsVisualizer rows={getResourceByType('Condition', groupedResources)} />
</PatientRecordElement>
 <PatientRecordElement
       resourceType={'Observation'}
       resourceCount={getResourceByType('Observation', groupedResources).length}
>
            <ObservationsVisualizer rows={getResourceByType('Observation', groupedResources)} />
</PatientRecordElement>
....
```

blangley28: The reason it looks so bad is that the group by function returns a map so doing `groupedResources.get(resourceType)` returns `DomainResource[] | undefined`.  This is unavoidable but I don't like the fact we have to keep repeating the resource type over and over again

dehall (approved):

radamson (approved):

PR #41, opened 2020-04-10
Expect MedicationRequests to be status "completed"

dehall:
Currently there is a discrepancy in the app when accepting a recommendation -- if the recommendation is a ServiceRequest, it reloads and the same state is still active though no longer actionable, but if the recommendation is a MedicationRequest, it advances through to the next state, even though that MedicationRequest hasn't been completed yet.

As it turns out, we had logic in there that ignored the "status" field for MedicationRequest, because at one point it seemed like there were multiple possible statuses for a MedicationRequest to indicate completion: "completed" and "stopped". As it turns out, "stopped" really means "stopped before completion" so we don't want to interpret that the same as "completed". Synthea uses "stopped" in its MedicationRequest resources but really it should be "completed" since the intent is that the entire course of treatment was done. (Thanks to May Terry for pointing that out in a review of some sample records)

This change checks for "completed" status again, and adds a helper function to determine if a resource is completed, which we can easily expand if needed.
I also updated demo patient Maureen's 2 medications to be completed instead of stopped, although that shouldn't really make a difference. 
Finally 2 unit tests had to be updated because they were expecting the patient to advance to the next state even though the relevant MedicationRequest is "active".

One other thing to be aware of - PATHWAYS-221 isn't one we've talked about in detail but I think it's something we'll need eventually. That task will add some way of manually advancing to the next state (for example, if a medication status was "stopped" but the oncologist wants to move forward with the pathway instead of cancelling it)

### DIFF
--- a/src/engine/output-results.ts
+++ b/src/engine/output-results.ts
@@ -208,21 +208,29 @@ function formatDocumentation(
 }
 
 /**
+ * Helper function to determine whether the action represented by
+ * the given resource has been "completed", and the pathway execution should advance.
+ * @param resource - the FHIR resource for a given action
+ * @return boolean as to whether this resource is complete
+ */
+function isComplete(resource: DocumentationResource): boolean {
+  // placeholder for more complex logic if needed.
+  // as of today MedicationRequest and ServiceRequest should both be "completed"
+  // (MedicationRequest can also be "stopped" though that indicates
+  // "Actions implied by the prescription are to be permanently halted, **before all of them occurred.**")
+  return resource.status === 'completed';
+}
+
+/**
  * Helper function to select the transition state
- * This function is needed because MedicationRequests can have multiple
- * different statuses to indiciate complete
  * @param resource - the resource returned by the CQL execution
  * @param currentState - the current state
  * @return the next state name or null
  */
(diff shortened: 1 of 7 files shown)


src/engine/output-results.ts
```diff
@@ -133,22 +133,30 @@ function formatDocumentation(
   return resource;
 }
 
+/**
+ * Helper function to determine whether the action represented by
+ * the given resource has been "completed", and the pathway execution should advance.
+ * @param resource - the FHIR resource for a given action
+ * @return boolean as to whether this resource is complete
```
radamson, line 214: I think you missed a dash here (but maybe not)
```suggestion
 * @return boolean - as to whether this resource is complete
```

radamson (approved):
Tested it out and the pathway no longer continues to the next state when selecting a medication.  Selecting a pathway provides an `active` state which previously was advancing the pathway.

Hopefully I'm interpreting this right 🙂 .

mdube7:
Test patients T0, T1, and triple_negative still have "stopped" as a status. The rest of this change looks good to me

mdube7 (commented):
Looks good to me

mdube7 (approved):

blangley28 (approved):

PR #42, opened 2020-04-13
Pathways assignment

danlee1025:
- Clicking the 'Select Pathway' button now adds an active `CarePlan` resource, which has the pathway name as it's `title` property, to the patient record.
    - After selecting a pathway, the button is disabled for the pathway until we implement the unselect pathway button.
    - Should work on demo and launcher endpoints.
- Added 'View Pathway' button
- How we're searching for the selected pathway
    - Looks for all active careplans that has a title that matches one of the pathways in the list

image
-
image
-
image

### DIFF
--- a/src/components/PathwaysList/PathwaysList.tsx
+++ b/src/components/PathwaysList/PathwaysList.tsx
@@ -230,6 +230,14 @@ const PathwaysListElement: FC<PathwaysListElementProps> = ({
             >
               Select Pathway
             </button>
+            <button
+              className={indexStyles.button}
+              onClick={(): void => {
+                callback(evaluatedPathway);
+              }}
+            >
+              View Pathway
+            </button>
           </div>
           <div className={styles.pathway}>
             <div style={{ height: '100%', overflow: 'scroll' }}>
--- a/src/components/PathwaysList/PathwaysList.tsx
+++ b/src/components/PathwaysList/PathwaysList.tsx
@@ -108,15 +108,15 @@ const PathwaysList: FC<PathwaysListProps> = ({ evaluatedPathways, callback, serv
   }
 
   const getSelectedPathway = (): string | undefined => {
-    // Get all active CarePlan resource titles
-    const carePlanTitles = (resources.filter(r => r.resourceType === 'CarePlan') as CarePlan[])
+    // Get all active CarePlan resource texts
+    const carePlanTexts = (resources.filter(r => r.resourceType === 'CarePlan') as CarePlan[])
       .filter(r => r.status === 'active')
-      .map(r => r.title);
+      .map(r => r.text?.div);
 
-    // Check to see if any of the pathway names are in carePlanTitles
+    // Check to see if any of the pathway names are in carePlanTexts
     const selectedPathway = evaluatedPathways
       .map(p => p.pathway.name)
-      .find(n => carePlanTitles.includes(n));
+      .find(n => carePlanTexts.some(text => text?.includes(n)));
 
     return selectedPathway;
   };
--- a/src/components/PathwaysList/PathwaysList.tsx
+++ b/src/components/PathwaysList/PathwaysList.tsx
@@ -12,7 +12,8 @@ import { IconProp } from '@fortawesome/fontawesome-svg-core';
 import { usePathwayContext } from 'components/PathwayProvider';
 import { evaluatePathwayCriteria } from 'engine';
 import { usePatientRecords } from 'components/PatientRecordsProvider';
-import { Resource } from 'fhir-objects';
+import { CarePlan, Patient } from 'fhir-objects';
+import { createCarePlan } from 'utils/fhirUtils';
 import {
   faPlay,
   faPlus,
@@ -21,6 +22,8 @@ import {
   faChevronDown,
   faCaretDown
 } from '@fortawesome/free-solid-svg-icons';
+import { usePatient } from 'components/PatientProvider';
+import { useFHIRClient } from 'components/FHIRClient';
 
 const useStyles = makeStyles(
   theme => ({
@@ -38,6 +41,7 @@ interface PathwaysListElementProps {
   evaluatedPathway: EvaluatedPathway;
   criteria?: CriteriaResult;
   callback: Function;
+  selected: boolean;
 }
 
 interface PathwaysListProps {
@@ -73,13 +77,15 @@ const PathwaysList: FC<PathwaysListProps> = ({ evaluatedPathways, callback, serv
         {criteria ? (
           criteria.map(c => {
             const evaluatedPathway = evaluatedPathways.find(p => p.pathway.name === c.pathwayName);
+            const pathwayName = evaluatedPathway?.pathway.name;
             if (evaluatedPathway)
               return (
                 <PathwaysListElement
                   evaluatedPathway={evaluatedPathway}
                   callback={callback}
                   criteria={c}
-                  key={evaluatedPathway.pathway.name}
+                  selected={pathwayName === selectedPathway}
+                  key={pathwayName}
                 />
               );
             else
@@ -92,6 +98,21 @@ const PathwaysList: FC<PathwaysListProps> = ({ evaluatedPathways, callback, serv
     );
   }
 
+  const getSelectedPathway = (): string | undefined => {
+    // Get all active CarePlan resource titles
+    const carePlanTitles = (resources.filter(r => r.resourceType === 'CarePlan') as CarePlan[])
+      .filter(r => r.status === 'active')
+      .map(r => r.title);
+
+    // Check to see if any of the pathway names are in carePlanTitles
+    const selectedPathway = evaluatedPathways
+      .map(p => p.pathway.name)
+      .find(n => carePlanTitles.includes(n));
+
+    return selectedPathway;
+  };
+
+  const selectedPathway = getSelectedPathway();
   return (
     <div className={styles.pathways_list}>
       {service.status === 'loading' ? (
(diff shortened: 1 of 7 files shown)


dehall:
2 early thoughts:
1) Clicking "Select Pathway" when there's already a selected pathway appears to be clickable but does nothing. From the code & description I see it should be disabled but it still has a hover color that makes it seem active.

2) It would be nice, though not mandatory, if the careplan was conformant to the US Core CarePlan profile: https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-careplan.html 
There's a lot we *could* put into the CarePlan eventually, like the selected actions etc, but for now just the basics is good.

danlee1025:
1. Do you think I should remove the "Select Pathway" button when a pathway is already selected?
2. I had it conforming to this https://www.hl7.org/fhir/careplan.html. I'll look to change it up.

dehall:
If we remove the button it might not be obvious to users why the button is gone. Maybe just make the disabled status a little more clear? Like make the button gray and no change of color on hover, and/or add a little alt text saying only one pathway may be selected at a time

danlee1025:
Rebased and removed WIP since design will mostly focus on Pathway view... any tweaks made in the design will be made in a future task.

danlee1025:
@dehall Button is grayed out when disabled and I added a small tooltip explaining why the button is disabled.

dehall:
Overall this looks good, the only sticking point to me is I'm now thinking do we really want to limit it to only one selected pathway per patient?
In terms of the technical aspects of the Explore Pathways screen, it's ultimately not a big difference either way. Thinking forward to selection/de-selection from the Pathway view though, I think it's going to be more complicated if that limit in place. We'd need some logic in place that checks for other assignments before allowing a new one.
In terms of the clinical aspect, obviously I'm not a doctor but it seems like there may be use cases where we want multiple active pathways. Or at least we may not want to explicitly block having multiple assigned pathways unless we have a specific reason to

danlee1025:
Good point, I don't think it will be too difficult to check for multiple selections.  I'm not sure why but I thought we said we only wanted one selection? Did you want me to change it to allow for multiple selections?

danlee1025:
Added functionality to allow selection of multiple pathways.

dehall (approved):

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -127,34 +155,55 @@ const PathwaysList: FC<PathwaysListProps> = ({ evaluatedPathways, callback, serv
...
+  const patient = usePatient().patient as Patient;
+  const client = useFHIRClient();
 
   const chevron: IconProp = isVisible ? faChevronUp : faChevronDown;
 
   function toggleVisible(): void {
     setIsVisible(!isVisible);
   }
 
+  const pathwayElementClass = selected
```
noranda: `clsx` actually has a way of handling conditionals. In this case you could write:
```js
const pathwayElementClass = clsx(
    selected && styles.selectedPathwayElement,
    selected && classes['selected-pathway-element'],
    !selected && styles.pathwayElement,
    !selected && classes['pathway-element']
  );
```

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -127,34 +155,55 @@ const PathwaysList: FC<PathwaysListProps> = ({ evaluatedPathways, callback, serv
...
+    ? clsx(styles.selectedPathwayElement, classes['selected-pathway-element'])
+    : clsx(styles.pathwayElement, classes['pathway-element']);
+  const titleClass = selected
```
noranda: This could be written as:
```js
const titleClass = clsx(
    styles.title,
    selected && classes['selected-title'],
    !selected && classes.title
  );
```

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -127,34 +155,55 @@ const PathwaysList: FC<PathwaysListProps> = ({ evaluatedPathways, callback, serv
...
+        className={titleClass}
         role={'listitem'}
         onClick={(e): void => {
           pathwayCtx.setEvaluatedPathway(evaluatedPathway, true);
           toggleVisible();
         }}
       >
         <div>{pathway.name}</div>
+        {selected && (
+          <div>
```
noranda: This div can be removed and these lines simplified to:
```js
{selected && <FontAwesomeIcon icon={faCheckCircle} />}
```

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -181,9 +230,27 @@ const PathwaysListElement: FC<PathwaysListElementProps> = ({
                 ))}
               </tbody>
             </table>
-            <button className={indexStyles.button} onClick={(): void => callback(evaluatedPathway)}>
+            <button
```
noranda, line 233: This button should have consistent styles to our other buttons. I would add `className={indexStyles.button}` to this button element and then add this to `src/styles/button.scss`:
```css
.button {
  ...

  &:disabled {
    background-color: lighten($blue, 20%);
    cursor: default;
  }
}
```

noranda (changes requested):
Just a few comments in the code, but overall this looks good.

The disabled button should eventually be replaced by the "unselect pathway" button but I assume that will happen in a later task. Right now there is no way to unselect a pathway.

I'm not sure how I feel about allowing multiple pathways to be selected. That may create confusion as to what the actual care plan is for the patient, but I might be missing something. I'd recommend running this by Jim or Zeshan to get their feedback.

src/components/PathwaysList/PathwaysList.tsx
noranda: I'm also a little confused why we've split up our styles this way. We seem to have styles in the component as well as a css module. Is there a reason why?

dehall:
Re: selecting multiple pathways, my take is we want to allow it unless there is a specific reason not to. We did talk about this a little while back, and if I recall correctly, one comment was that a patient may not be on multiple oncology pathways at the same time but they might be on pathways for other conditions, like diabetes, hypertension, etc.

src/components/PathwaysList/PathwaysList.tsx
danlee1025: I think `pathway-element` and `title` were styled in the component for theming so I added `selected-pathway-element` and `selected-title` to change the properties to stay consistent. Do you think I should move the newly added ones into the css module?

noranda:
> Re: selecting multiple pathways, my take is we want to allow it unless there is a specific reason not to. We did talk about this a little while back, and if I recall correctly, one comment was that a patient may not be on multiple oncology pathways at the same time but they might be on pathways for other conditions, like diabetes, hypertension, etc.

Ok I think this makes sense. Thanks for clarifying!

src/components/PathwaysList/PathwaysList.tsx
noranda: Yeah I think it's confusing to split up the styling like this. Theming is done from the `styles/theme.tsx` file. Do we do this in other files? Thoughts, @radamson or @dehall?

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -92,6 +105,21 @@ const PathwaysList: FC<PathwaysListProps> = ({ evaluatedPathways, callback, serv
     );
   }
 
+  const getSelectedPathways = (): string[] => {
+    // Get all active CarePlan resource texts
+    const carePlanTexts = (resources.filter(r => r.resourceType === 'CarePlan') as CarePlan[])
+      .filter(r => r.status === 'active')
+      .map(r => r.text?.div);
```
blangley28: Is there a reason you chose text here over title? I feel though title is more appropriate so instead of checking the text includes the pathway name you just check the title is the same as the pathway name. Both work but title seems more natural to me

blangley28: Just to clarify some more after reviewing the createCarePlan function - we should still leave the text explaining what it is but also setting the title will make for an easier comparison than checking the pathway name is included in the text

danlee1025: yea I actually had it checking the title at the beginning of this PR, but I changed the CarePlan to conform with USCore profile which required the text property. Since we were including the pathway name in the text property I decided to just check for it there. It's actually easier to check for it in the title so I'm not opposed to adding the title property back. Any preference @dehall?

dehall: Yeah that sounds good. Keep text as "assignment of patient to pathway {x}" and then title is just the pathway name

danlee1025: Cool, this change has been pushed.

src/components/PathwaysList/PathwaysList.tsx
danlee1025: @noranda So it looks like Dylan and Reece will defer to you. How do you suggest we consolidate the styles? Is `Node.tsx` doing something similar and should we consolidate it there also?

radamson: I see what's going on after staring at it for a bit.  It would be nice to consolidate the styles here, but I don't think we need to do it for every file in this PR.  Depending on how tricky it is I don't have a problem doing it here or keeping for a subsequent PR.

**Old way of doing styles:**
Most of our components are styled using separate SASS style sheets that are imported.

```typescript
import styles from './PathwaysList.module.scss'

...

<div className = {styles.pathwaysStyle} />
```

**New way of doing styles:**
The new way of doing styles involves using JSS ~~in the same component file~~, instead of SASS.

In https://github.com/mcode/pathways/pull/26 we started using material-ui which offers some nice utilities to help with styles and theming (which is really just a thin layer over JSS under the hood).

The nice thing here is it makes it really easy to apply the app theme through `makeStyles`.

Also, @noranda, looking at https://github.com/mcode/pathways/pull/26/files I think you're the one who originally added it this way (i.e. using both `classes` and `styles`) 🙂 .

noranda: > Also, @noranda, looking at https://github.com/mcode/pathways/pull/26/files I think you're the one who originally added it this way (i.e. using both `classes` and `styles`) 🙂 .

Ah yes. I think I was under the impression that we would slowly convert to this way of using styles, although using JSS with `makeStyles` can and should be done in a separate file.

I agree, let's add a task to do the conversion. I'm okay leaving this PR as is if we agree to fix this in a future task.

noranda (approved):
Approved with the caveat that we consolidate the styles in a future task.

src/components/PathwaysList/PathwaysList.tsx
radamson: @noranda sounds good!  I created PATHWAYS-237 to move all the styles over and put in the description that all the JSS/makeStyles should be done in separate files from the component.  Good catch on this; it's an important task.

PR #43, opened 2020-04-22
Mcode sidebar info

KeeyanGhoreshi:
Adds the mcode info to the sidebar. Thinking about changing the first column text to look better, thoughts?

### DIFF
--- a/src/components/PatientRecord/PatientRecord.tsx
+++ b/src/components/PatientRecord/PatientRecord.tsx
@@ -26,9 +26,11 @@ import { IconProp } from '@fortawesome/fontawesome-svg-core';
 import { CriteriaResult } from 'pathways-model';
 import { usePathwayContext } from 'components/PathwayProvider';
 import { evaluatePathwayCriteria } from 'engine';
+import { McodeElements } from 'mcode';
 
 const resourceTypes = [
   'Pathway',
+  'Mcode',
   'Patient',
   'Condition',
   'Observation',
@@ -124,10 +126,12 @@ const PatientRecord: FC<PatientRecordProps> = ({ headerElement }) => {
 };
 
 const PatientRecordElement: FC<PatientRecordElementProps> = ({ resourceType, resources }) => {
+  console.log(resourceType);
+  console.log(resources);
   const [isExpanded, setIsExpanded] = useState<boolean>(false);
 
   const chevron: IconProp = isExpanded ? faChevronUp : faChevronDown;
-  const resourceCount: string = !['Patient', 'Pathway'].includes(resourceType)
+  const resourceCount: string = !['Patient', 'Pathway', 'Mcode'].includes(resourceType)
     ? `(${resources.length})`
     : '';
 
@@ -155,6 +159,7 @@ const Visualizer: FC<VisualizerProps> = ({ resourceType, resourcesByType }) => {
   const patient = usePatient().patient as fhir.Patient;
 
   if (resourceType === 'Pathway') return <PathwayVisualizer />;
+  else if (resourceType === 'Mcode') return <McodeVisualizer />;
   else if (resourceType === 'Patient') return <PatientVisualizer patient={patient} />;
   else if (resourceType === 'Condition') return <ConditionsVisualizer rows={resourcesByType} />;
   else if (resourceType === 'Observation') return <ObservationsVisualizer rows={resourcesByType} />;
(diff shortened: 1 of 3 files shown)


src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -124,10 +126,12 @@ const PatientRecord: FC<PatientRecordProps> = ({ headerElement }) => {
 };
 
 const PatientRecordElement: FC<PatientRecordElementProps> = ({ resourceType, resources }) => {
+  console.log(resourceType);
+  console.log(resources);
```
blangley28: These log statements should be removed

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -206,4 +211,28 @@ const PathwayVisualizer: FC = () => {
   );
 };
 
+const McodeVisualizer: FC = () => {
+  const mcode = usePatientRecords().mcodeRecords;
+  console.log(mcode);
```
blangley28: Same with this log statement

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -206,4 +211,28 @@ const PathwayVisualizer: FC = () => {
   );
 };
 
+const McodeVisualizer: FC = () => {
+  const mcode = usePatientRecords().mcodeRecords;
```
blangley28: We should be consistent with how we use specific parts of the context. I have seen this as well as 
`const {element} = useProvider()`

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -206,4 +211,28 @@ const PathwayVisualizer: FC = () => {
...
+  for (const key in mcode) {
+    keysArray.push(key as mcodeKey);
+  }
+  return (
+    <table>
+      <tbody>
+        {keysArray.map(key => {
+          return (
+            <tr key={key}>
+              <td>{key}</td>
```
blangley28, line 227: We should consider formatting the key here. Without introducing any other data a simple camel case convertor could work - capitalize the first character and then add a space before each capital letter

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -206,4 +211,28 @@ const PathwayVisualizer: FC = () => {
...
+              <td>{mcode[key] ? mcode[key] : '-'}</td>
```
blangley28: Not sure if it falls under this task or not but instead of `-` for missing data it should be the missing data link. Perhaps make a backlog item for it?

KeeyanGhoreshi: I think it'll have to be a backlog task because there's no way to allow the user to hand select missing data beyond just selecting a transition option.

blangley28: Sounds good to me. Created https://jira.mitre.org/browse/PATHWAYS-241

blangley28 (approved):

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -155,6 +157,7 @@ const Visualizer: FC<VisualizerProps> = ({ resourceType, resourcesByType }) => {
   const patient = usePatient().patient as fhir.Patient;
 
   if (resourceType === 'Pathway') return <PathwayVisualizer />;
+  else if (resourceType === 'Mcode') return <McodeVisualizer />;
```
radamson: The way `Mcode` and `Pathway` are handled in this file is confusing and I had to look at it for a bit to realize what was going on.  It suggests that they are `resourceTypes` just like FHIR resources.  I know previous PRs have discussed ways to reduce the number of `if`s in this file, but even just the naming is confusing to understanding what's going on.

I think we should at least rename `resourceTypes` on line 31 to `recordSections`.

src/components/PatientRecord/PatientRecord.tsx
```diff
@@ -206,4 +209,29 @@ const PathwayVisualizer: FC = () => {
...
+  type mcodeKey = keyof McodeElements & string;
+  const keysArray: mcodeKey[] = [];
+  for (const key in mcodeRecords) {
+    keysArray.push(key as mcodeKey);
+  }
+  return (
+    <table>
+      <tbody>
+        {keysArray.map(key => {
+          const result = key.replace(/([A-Z]+)*([A-Z][a-z])/g, '$1 $2');
```
radamson: I'm not a regex expert and regexes can be hard to read so a comment would be helpful here.

Is there any reason we can't use a simpler regex like `/([A-Z])/g`?  the current one looks like it has an undefined first match baed on some quick testing.

Also this might be hard, but right now HER2 is displayed as `Her2 Receptor` and I think it should be `HER2 Receptor`.  Optimally we would have display values for these mcode elements set.

radamson:
Screenshot showing expanded MCODE section

Screen Shot 2020-04-27 at 1 40 54 PM

src/components/PatientRecord/PatientRecord.tsx
dehall: Not entirely sure if this is legal from a TypeScript or style standpoint, but why not just make the field key directly match what we want the display value to be? Otherwise we need a brittle regex or a manual mapping of keys to field names.

i.e., in mcode.d.ts:
```ts
  export interface McodeElements {
    'Primary Cancer'?: string;
    'Laterality'?: string;
    'Tumor Category'?: string;
    'Node Category'?: string;
```

blangley28: Just did a quick test and this works. I think its the best solution because as you mentioned it doesn't require regex nor manual mapping

radamson: I like this solution too

blangley28: I updated this on my branch and then made this branch use the keys

radamson (approved):

PR #44, opened 2020-04-23
mCODE Data + Report Modal

blangley28:
Creates a new CQL file to grab mCODE data from the patient record and then make it available in the patient provider. Then it uses this data to populate the report modal.

Screen Shot 2020-04-23 at 2 35 14 PM

Putting this PR up now for comments but Keeyan's branch should be merged into this first.
Note: eslint fails but this is fixed by Keeyan's PR

### DIFF
--- a/public/static/cql/mCODE.cql
+++ b/public/static/cql/mCODE.cql
@@ -12,35 +12,34 @@ code "Primary cancer code": '64572001' from "SNOMEDCT" display 'Disease (disorde
 code "TNM clinical primary tumor code": '21905-5' from "LOINC" display 'Primary tumor.clinical [Class] Cancer'
 code "TNM clinical regional nodes code": '21906-3' from "LOINC" display 'Regional lymph nodes.clinical [Class] Cancer'
 code "TNM clinical distant metastases code": '21907-1' from "LOINC" display 'Distant metastases.clinical [Class] Cancer'
-code "Tumor marker test code": 'laboratory' from "OBSCAT"
 code "Estrogen receptor code": '85337-4' from "LOINC" display 'Estrogen receptor Ag [Presence] in Breast cancer specimen by Immune stain'
 code "Progesterone receptor code": '85339-0' from "LOINC" display 'Progesterone receptor Ag [Presence] in Breast cancer specimen by Immune stain'
 code "HER2 receptor code": '85319-2' from "LOINC" display 'HER2 [Presence] in Breast cancer specimen by Immune stain'
 
 // selected cancer codes used within Synthea
 // not the entire list supported by PrimaryCancerConditionVS
-code "11p partial monosomy syndrome 4135001 code": '4135001' from SNOMEDCT display '11p partial monosomy syndrome'
-code "Orbital lymphoma 13048006 code": '13048006' from SNOMEDCT display 'Orbital lymphoma'
-code "Delta heavy chain disease 20224008 code": '20224008' from SNOMEDCT display 'Delta heavy chain disease'
-code "Malignant neoplasm of breast 254837009 code": '254837009' from SNOMEDCT display 'Malignant neoplasm of breast'
-code "Primary malignant neoplasm of colon 93761005 code": '93761005' from SNOMEDCT display 'Primary malignant neoplasm of colon'
-code "Secondary malignant neoplasm of colon 94260004 code": '94260004' from SNOMEDCT display 'Secondary malignant neoplasm of colon'
-code "Carcinoma in situ of prostate (disorder) 92691004 code": '92691004' from SNOMEDCT display 'Carcinoma in situ of prostate (disorder)'
-code "Small cell carcinoma of lung (disorder) 254632001 code": '254632001' from SNOMEDCT display 'Small cell carcinoma of lung (disorder)'
-code "Non-small cell lung cancer (disorder) 254637007 code": '254637007' from SNOMEDCT display 'Non-small cell lung cancer (disorder)'
+code "11p partial monosomy syndrome code": '4135001' from SNOMEDCT display '11p partial monosomy syndrome'
+code "Orbital lymphoma code": '13048006' from SNOMEDCT display 'Orbital lymphoma'
+code "Delta heavy chain disease code": '20224008' from SNOMEDCT display 'Delta heavy chain disease'
+code "Malignant neoplasm of breast code": '254837009' from SNOMEDCT display 'Malignant neoplasm of breast'
+code "Primary malignant neoplasm of colon code": '93761005' from SNOMEDCT display 'Primary malignant neoplasm of colon'
+code "Secondary malignant neoplasm of colon code": '94260004' from SNOMEDCT display 'Secondary malignant neoplasm of colon'
+code "Carcinoma in situ of prostate (disorder) code": '92691004' from SNOMEDCT display 'Carcinoma in situ of prostate (disorder)'
+code "Small cell carcinoma of lung (disorder) code": '254632001' from SNOMEDCT display 'Small cell carcinoma of lung (disorder)'
+code "Non-small cell lung cancer (disorder) code": '254637007' from SNOMEDCT display 'Non-small cell lung cancer (disorder)'
 
 // Category code as concepts
 concept "Primary cancer": { "Primary cancer code" }
 concept "PrimaryCancersVS": {
-    "11p partial monosomy syndrome 4135001 code",
-    "Orbital lymphoma 13048006 code",
-    "Delta heavy chain disease 20224008 code",
-    "Malignant neoplasm of breast 254837009 code",
-    "Primary malignant neoplasm of colon 93761005 code",
-    "Secondary malignant neoplasm of colon 94260004 code",
-    "Carcinoma in situ of prostate (disorder) 92691004 code",
-    "Small cell carcinoma of lung (disorder) 254632001 code",
-    "Non-small cell lung cancer (disorder) 254637007 code":
+    "11p partial monosomy syndrome code",
+    "Orbital lymphoma code",
+    "Delta heavy chain disease code",
+    "Malignant neoplasm of breast code",
+    "Primary malignant neoplasm of colon code",
+    "Secondary malignant neoplasm of colon code",
+    "Carcinoma in situ of prostate (disorder) code",
+    "Small cell carcinoma of lung (disorder) code",
+    "Non-small cell lung cancer (disorder) code":
 }
 
 context Patient
--- a/public/static/cql/mCODE.cql
+++ b/public/static/cql/mCODE.cql
@@ -4,9 +4,99 @@ using FHIR version '4.0.0'
 
 // CODESYSTEMS
 codesystem "SNOMEDCT": 'http://snomed.info/sct'
+codesystem "LOINC": 'http://loinc.org'
+codesystem "OBSCAT": 'http://terminology.hl7.org/CodeSystem/observation-category'
+
+// FHIR CODES
+code "Primary cancer code": '64572001' from "SNOMEDCT" display 'Disease'
+code "TNM clinical primary tumor code": '21905-5' from "LOINC" display 'Primary tumor.clinical [Class] Cancer'
+code "TNM clinical regional nodes code": '21906-3' from "LOINC" display 'Regional lymph nodes.clinical [Class] Cancer'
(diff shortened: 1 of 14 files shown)


public/static/cql/mCODE.cql
```diff
@@ -4,9 +4,121 @@ using FHIR version '4.0.0'
...
 codesystem "SNOMEDCT": 'http://snomed.info/sct'
+codesystem "LOINC": 'http://loinc.org'
+codesystem "OBSCAT": 'http://terminology.hl7.org/CodeSystem/observation-category'
+
+// FHIR CODES
+code "Primary cancer code": '64572001' from "SNOMEDCT" display 'Disease (disorder)'
+code "TNM clinical primary tumor code": '21905-5' from "LOINC" display 'Primary tumor.clinical [Class] Cancer'
+code "TNM clinical regional nodes code": '21906-3' from "LOINC" display 'Regional lymph nodes.clinical [Class] Cancer'
+code "TNM clinical distant metastases code": '21907-1' from "LOINC" display 'Distant metastases.clinical [Class] Cancer'
+code "Tumor marker test code": 'laboratory' from "OBSCAT"
```
radamson: Should we have the display here like with the other codes?

public/static/cql/mCODE.cql
```diff
@@ -4,9 +4,121 @@ using FHIR version '4.0.0'
...
+code "Estrogen receptor code": '85337-4' from "LOINC" display 'Estrogen receptor Ag [Presence] in Breast cancer specimen by Immune stain'
+code "Progesterone receptor code": '85339-0' from "LOINC" display 'Progesterone receptor Ag [Presence] in Breast cancer specimen by Immune stain'
+code "HER2 receptor code": '85319-2' from "LOINC" display 'HER2 [Presence] in Breast cancer specimen by Immune stain'
+
+// selected cancer codes used within Synthea
+// not the entire list supported by PrimaryCancerConditionVS
+code "11p partial monosomy syndrome 4135001 code": '4135001' from SNOMEDCT display '11p partial monosomy syndrome'
```
radamson: Is there a reason these are declared with the code in the name?

i.e.

`11p partial monosomy syndrome 4135001 code`

vs

`11p partial monosomy syndrome`

src/components/App.tsx
```diff
@@ -42,14 +47,61 @@ const App: FC<AppProps> = ({ demoId }) => {
...
+      .then(elm => {
+        let elmResults: ElmResults = {
+          patientResults: {}
+        };
+        elmResults = executeElm(bundle, elm);
+        const patientIds = Object.keys(elmResults.patientResults);
+        const mcodeData = elmResults.patientResults[patientIds[0]];
+
+        const mcodeElements: McodeElements = {
+          primaryCancer: mcodeData['Primary Cancer Condition Code'][0] ?? undefined,
```
radamson: feels weird to be setting something to undefined while defining it.  Does null work here?

src/components/App.tsx
```diff
@@ -42,14 +47,61 @@ const App: FC<AppProps> = ({ demoId }) => {
...
+    // Create a Bundle for the CQL engine
+    const bundle = {
+      resourceType: 'Bundle',
+      type: 'searchset',
+      entry: resources.map((r: DomainResource) => ({ resource: r }))
+    };
+    getFixture('mCODE.cql')
+      .then(cql => convertBasicCQL(cql))
+      .then(elm => {
+        let elmResults: ElmResults = {
```
radamson: This variable initializer is redundant.  You can remove it and just use the definition on line 63.

src/components/App.tsx
```diff
@@ -42,14 +47,61 @@ const App: FC<AppProps> = ({ demoId }) => {
...
+          patientResults: {}
+        };
+        elmResults = executeElm(bundle, elm);
+        const patientIds = Object.keys(elmResults.patientResults);
```
radamson: I suppose any resources could be passed into this function, but you should also be able to just get the id out of the `patient.id` state.  I'm not sure if there is an advantage to doing it one way or another, but at first glance it seems advantageous to grab it from the already available source.

src/utils/fhirUtils.ts
```diff
@@ -211,3 +212,13 @@ export function createCarePlan(title: string, patient: Patient): CarePlan {
     subject: { reference: `Patient/${patient.id}` }
   };
 }
+
+export function getTNM(mcodeElements: McodeElements): string {
+  let tnm = '';
```
radamson: clearly this works, but its a lot of `+=` and `??` to get the whitespace right.  Could probably use join here (and avoid a trailing whitespace if there is no `metastatsesCategory`)

I didn't actually test this so I might have a syntax error.

```typescript
return [ mcodeElements.tumorCategory?.split(' ')[0],
              mcodeElements.nodeCategory?.split(' ')[0],
              mcodeElements.metastasesCategory?.split(' ')[0]
           ].join(' ')

File: src/components/ReportModal/ReportModal.tsx [hunk 7]
```diff
@@ -21,12 +24,12 @@ const ReportModal: FC<ReportModalInterface> = ({ onConfirm, onDecline }) => {
...
   const observationField = [
-    { name: 'Primary Cancer', value: undefined },
-    { name: 'Laterality', value: undefined },
-    { name: 'Clinical TNM', value: undefined },
-    { name: 'Estrogen Receptor', value: undefined },
-    { name: 'Progesterone Receptor', value: undefined },
-    { name: 'HER2 Receptor', value: undefined }
+    { name: 'Primary Cancer', value: mcodeRecords.primaryCancer ?? 'Unknown' },
+    { name: 'Laterality', value: mcodeRecords.laterality ?? 'Unknown' },
+    { name: 'Clinical TNM', value: getTNM(mcodeRecords) },
```
radamson, line 29: should this one also have `?? 'Unknown'` ?

radamson (commented):
Tested it out and looks good.  I didn't check for linting in this PR.  I had a couple of suggestions and questions to help confirm my understanding.

I'm always learning something when CQL is involved 🙂

public/static/cql/mCODE.cql
blangley28: @dehall I think you wrote these originally for the criteria? All of the `code`s include "code" at the end but as for including the numerical code I'm not sure why

src/components/App.tsx
blangley28: I went with undefined because the types are declared as optional. Making them null is a possibility with updating the type

src/components/App.tsx
blangley28: Let me give it a test but IIRC without explicitly defining the `patientResults` element inside `elmResults` I was unable to read the data correctly. I copied this from somewhere else in the engine and also thought it was peculiar but deleting it failed.

src/components/App.tsx
radamson: eh, we can just leave it undefined then I forgot about changing the type.

src/components/App.tsx
radamson: my IDE was picking it up  ¯\_(ツ)_/¯

src/utils/fhirUtils.ts
blangley28: It would avoid the trailing whitespace but also introduce a double space if `nodeCategory` is undefined. At the end of the day neither are a big deal and I think this suggestion is cleaner so I'll go with it

src/components/ReportModal/ReportModal.tsx
blangley28: No, `getTNM` always returns a string. However if all three TNM values are undefined the result will be "   " so I will add a check to the method so it returns "Unknown" if all 3 are undefined

public/static/cql/mCODE.cql
dehall: I copied them into the mCODE_Library.cql file from an old version of the CQL on our internal repo: https://gitlab.mitre.org/mcode/pathways/-/blob/master/src/model/valuesets/PrimaryCancerVS.cql

That said I'm fine with whatever format we want to use for the code names, because we should only be referencing them internally in the definition of a concept.

src/components/App.tsx
blangley28: Just gave this a try but I think it will cause more headache than gain since `patient` can be null and `id` can be undefined

public/static/cql/mCODE.cql
blangley28: Removed since this code isn't actually used

public/static/cql/mCODE.cql
blangley28: I removed the numerical code values for consistency with the rest of the file.
Ie: "11p partial monosomy syndrome 4135001 code" --> "11p partial monosomy syndrome code"

public/static/cql/mCODE.cql
```diff
@@ -4,9 +4,120 @@ using FHIR version '4.0.0'
 
 // CODESYSTEMS
 codesystem "SNOMEDCT": 'http://snomed.info/sct'
+codesystem "LOINC": 'http://loinc.org'
+codesystem "OBSCAT": 'http://terminology.hl7.org/CodeSystem/observation-category'
```
radamson: It doesn't look like this has any uses and can be removed now that the code using it has been removed.

blangley28: Good catch. Removed

src/components/App.tsx
```diff
@@ -42,14 +47,58 @@ const App: FC<AppProps> = ({ demoId }) => {
...
 
+  const setMcodeRecords = useCallback((resources: DomainResource[]): void => {
+    // Create a Bundle for the CQL engine
+    const bundle = {
+      resourceType: 'Bundle',
+      type: 'searchset',
+      entry: resources.map((r: DomainResource) => ({ resource: r }))
+    };
+    getFixture('mCODE.cql')
+      .then(cql => convertBasicCQL(cql))
```
dehall: Instead of converting the CQL to ELM on every app load can we pre-convert it and store the ELM somewhere? Additionally maybe we should look into extracting the mCODE CQL library into its own repository again, though that's not part of this task

src/components/App.tsx
```diff
@@ -42,14 +47,58 @@ const App: FC<AppProps> = ({ demoId }) => {
...
+      .then(elm => {
+        let elmResults = executeElm(bundle, elm);
+        const patientIds = Object.keys(elmResults.patientResults);
+        const mcodeData = elmResults.patientResults[patientIds[0]];
+
+        const mcodeElements: McodeElements = {
+          primaryCancer: mcodeData['Primary Cancer Condition Code'][0] ?? undefined,
+          laterality: mcodeData['Primary Cancer Condition Body Location Code'][0] ?? undefined,
+          tumorCategory:
+            mcodeData['TNM Clinical Primary Tumor Category Data Value (T Category)'][0] ??
```
dehall: To make this a little more robust, instead of just fetching the `text` field in the referenced CQL could these also look at, for example, the `code.display` field? We don't have to account for all the many possibilities here, like interpreting codes if both text and code.display are empty (which is totally legal), but test data may not always include the text field. Maybe it's a helper function in the CQL or in the code here but it shouldn't be too complex

src/components/App.tsx
```diff
@@ -42,14 +47,58 @@ const App: FC<AppProps> = ({ demoId }) => {
     setEvaluatePath(true);
   }, []);
 
+  const setMcodeRecords = useCallback((resources: DomainResource[]): void => {
+    // Create a Bundle for the CQL engine
```
dehall, line 50: We do this in so many places I wonder if it makes more sense to just make the ELM executor take in an array instead of a bundle
Edit: this would be more appropriate as a very quick 1-pt task instead of piling on unrelated changes here

dehall (commented):
Looks good -- main thing to me is preconverting the CQL to ELM

src/components/App.tsx
radamson: That's a good point @dehall.  From my experience, and without any evidence to back it up, display is more common (and doesn't surprise me because it is taken right from the system).

blangley28: I'll update it to read `code[0].coding[0].display.value`

src/components/App.tsx
blangley28: Added pre-converted elm. I agree - the task already exists in the backlog: https://jira.mitre.org/browse/PATHWAYS-117

src/components/App.tsx
blangley28: https://jira.mitre.org/browse/PATHWAYS-242

radamson (approved):

dehall (approved):

PR #45, opened 2020-04-27
Fix Node Status Bug

blangley28:
There was a bug where action nodes were incorrectly showing status for nodes not on the current path. Previously the Chemotherapy nodes for the HER2 pathway had documentation when expanded but did not mark themselves as complete on the status icon.
Screen Shot 2020-04-27 at 2 29 29 PM
Note: Use demo patient Maryam


danlee1025 (approved):
LGTM

dehall (approved):

PR #46, opened 2020-04-30
make scrollables modify height on rerender

KeeyanGhoreshi:
Basically, a specific `useEffect` hook wouldn't fire on rerender and the `height` style simply wouldn't be set on the div, so it would just stretch to fit all the content, making the body scroll instead, leaving the header behind.  

The way I fixed it was to make the `useEffect` run every render by removing the dependency list.  I did the same thing for the `PathwayList` since it would leave the header behind as well when the body scrolled.

There could a more elegant solution using CSS, since our real problem is that we want the divs that contain the content to have a max-height equal to the size of the window minus the size of the header.  If the header was a consistent size (perhaps a function of vh) then the containers could also have a consistent max-height.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -158,10 +158,11 @@ const App: FC<AppProps> = ({ demoId }) => {
 
   // Set the height of the graph container
   useEffect(() => {
-    if (graphContainerElement?.current && headerElement?.current)
+    if (graphContainerElement?.current && headerElement?.current) {
       graphContainerElement.current.style.height =
         window.innerHeight - headerElement.current.clientHeight + 'px';
-  }, [selectPathway]);
+    }
+  });
 
   function setEvaluatedPathwayCallback(
     value: EvaluatedPathway | null,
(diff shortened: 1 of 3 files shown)


blangley28 (approved):
LGTM and does indeed fix the bug

blangley28 (changes requested):
So when I was reviewing this I noticed a similar bug to the one I just fixed in the PathwaysBuilder. It was an easy change so I did it super quick branching from master and it actually fixes this bug. Essentially what was happening is the PatientView component is defined with the App component. So my belief is that whenever the App re-renders the PatientView is unmounted and then a brand new one is mounted. When this occurs that useEffect hook was not running. By moving the PatientView component out of the App component it now fixes this problem. I think we should go with this solution since it doesn't require removing the dependency list on the useEffect hook and ends up killing two birds with one stone. I pushed the changes to fix-unmounting-bug which I can put up for a PR

KeeyanGhoreshi:
I'll reinstate the dependency but keep the scrolling behavior on PathwaysList

blangley28 (approved):

dehall (approved):

PR #49, opened 2020-05-22
Modified declined nodes to be more visually distinguishable from acce…

ms-k1ngk0ng:
…pted nodes.

### DIFF
--- a/src/components/ExpandedNode/ExpandedNode.tsx
+++ b/src/components/ExpandedNode/ExpandedNode.tsx
@@ -136,14 +136,16 @@ const StatusField: FC<StatusFieldProps> = ({ documentation }) => {
   }
   const status = documentation.status;
   const rawDate = documentation.resource?.meta?.lastUpdated;
-  if (rawDate)
+  if (rawDate) {
+    const descript = new Date(rawDate).toLocaleString('en-us');
     return (
       <ExpandedNodeField
         key="Status"
         title={status}
-        description={new Date(rawDate).toLocaleString('en-us')}
+        description={status === 'Declined' ? descript.concat(' by Dr. Example') : descript}
       />
     );
+  }
   return null;
 };
 
--- a/src/components/ExpandedNode/ExpandedNode.tsx
+++ b/src/components/ExpandedNode/ExpandedNode.tsx
@@ -144,8 +144,8 @@ const StatusField: FC<StatusFieldProps> = ({ documentation, isAccepted }) => {
     return (
       <ExpandedNodeField
         key="Status"
-        title={isAccepted === false ? 'Declined' : status}
-        description={isAccepted === false ? date.concat(' by Dr. Example') : date}
+        title={isAccepted ? status.charAt(0).toUpperCase() + status.slice(1) : 'Declined'}
+        description={isAccepted ? date : date.concat(' by Dr. Example')}
       />
     );
   }
@@ -362,16 +362,18 @@ const ExpandedNodeMemo: FC<ExpandedNodeMemoProps> = memo(
         </table>
         {isActionable && isGuidance && (
           <form className={styles.commentsForm}>
-            <label>Comments:</label>
-            <button
-              className={`${indexStyles.button} ${styles.defaultTextButton}`}
-              onClick={(e): void => {
-                e.preventDefault();
-                if (!comments.includes(defaultText)) setComments(comments + defaultText);
-              }}
-            >
-              Use Default Text
-            </button>
+            <div>
+              <label>Comments:</label>
+              <button
+                className={`${indexStyles.button} ${styles.defaultTextButton}`}
+                onClick={(e): void => {
+                  e.preventDefault();
+                  if (!comments.includes(defaultText)) setComments(comments + defaultText);
+                }}
+              >
+                Use Default Text
+              </button>
+            </div>
             <textarea
               className={styles.comments}
               value={comments}
--- a/src/components/ExpandedNode/ExpandedNode.tsx
+++ b/src/components/ExpandedNode/ExpandedNode.tsx
@@ -31,10 +31,11 @@ interface ExpandedNodeProps {
   isActionable: boolean;
   isGuidance: boolean;
   documentation: DocumentationResource | undefined;
+  isAccepted: boolean | null;
 }
 
 const ExpandedNode: FC<ExpandedNodeProps> = memo(
-  ({ pathwayState, isActionable, isGuidance, documentation }) => {
+  ({ pathwayState, isActionable, isGuidance, documentation, isAccepted }) => {
     const { note, setNote } = useNote();
     const [showReport, setShowReport] = useState<boolean>(false);
     const { patientRecords, setPatientRecords } = usePatientRecords();
@@ -100,6 +101,7 @@ const ExpandedNode: FC<ExpandedNodeProps> = memo(
             });
             setShowReport(true);
           }}
+          isAccepted={isAccepted}
         />
         {showReport && (
           <ReportModal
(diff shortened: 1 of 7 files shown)


src/components/Node/Node.module.scss
```diff
@@ -28,6 +28,10 @@
     background-color: $red;
   }
 
+  &.declined {
+    background-color: $black;
```
blangley28: Taking a look at the design I don't think the background color is black. From a color picker it looks like `#404040`. @noranda is this the correct color?

blangley28 (changes requested):
Woohoo first PR! Looks good but I think the color is different. You probably didn't notice this in the design but there is also a difference in the text of the expanded node. Instead of saying "current: {time}" it should say "declined: {time} by {doctor}". I think the doctor should come from the `UserProvider` context.

Also comment on the icon - it doesn't look inline with the text. I'm not sure which is correctly centered but it just seems off to me.

dehall:
We would want the name of the declining doctor to be the user that clicked the decline button, which isn't guaranteed to be the current user. (i.e., someone else could load up the patient/pathway later and want to see more about a declined action) So that would probably come from the note not the UserProvider, and that might be a larger change than intended. 

For now if we can easily change the text to say "declined" instead of "status" without too much trouble, let's do that.

blangley28:
> So that would probably come from the note not the UserProvider, and that might be a larger change than intended.

Yea you are totally right. I did a quick look to determine how hard this would be:
The note does include the provider and is written to the DocumentReference which is put in the patient records. The correct DocumentReference is also returned as the `documentation` when rendering an ExpandedNode (see `renderGuidance()` in `ExpandedNode.tsx`). This makes things pretty easy. To get the provider we would probably need to write a regex to pull it out of the note. I also noticed in demo mode we don't have a user set in demo mode so we should probably add that too.

ms-k1ngk0ng:
Screen Shot 2020-05-27 at 4 13 18 PM
Just incorporated your feedback @blangley28. Had a difficult time adjusting the icons and label, but should be good now - let me know what you think!

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -136,14 +136,16 @@ const StatusField: FC<StatusFieldProps> = ({ documentation }) => {
   }
   const status = documentation.status;
   const rawDate = documentation.resource?.meta?.lastUpdated;
-  if (rawDate)
+  if (rawDate) {
+    const descript = new Date(rawDate).toLocaleString('en-us');
```
blangley28: Can we rename this to date just so its clear down below
```suggestion
    const date = new Date(rawDate).toLocaleString('en-us');
```

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -136,14 +136,16 @@ const StatusField: FC<StatusFieldProps> = ({ documentation }) => {
...
     return (
       <ExpandedNodeField
         key="Status"
         title={status}
-        description={new Date(rawDate).toLocaleString('en-us')}
+        description={status === 'Declined' ? descript.concat(' by Dr. Example') : descript}
```
blangley28:
```suggestion
        description={status === 'Declined' ? date.concat(' by Dr. Example') : date}
```

src/components/Node/Node.tsx
```diff
@@ -89,6 +89,11 @@ const Node: FC<NodeProps & { ref: Ref<HTMLDivElement> }> = memo(
       const isAccepted = wasActionTaken
         ? documentation?.resourceType !== 'DocumentReference'
         : null;
+      if (isAccepted === false) {
+        topLevelClasses.push(styles.declined);
+        if (documentation) documentation.status = 'Declined';
```
blangley28: Is there a way to do this without changing the status. I don't _think_ we are writing this back anywhere but just in case we should avoid it. 'Declined' is not actually a status for the resources we are using for documentation. 
The two solutions I can think of are
1. Passing the `isAccepted` flag down to the `ExpandedNode` (as a prop)
2. A decline is done through a DocumentReference with an attached note. You can check the status of that note for accepted/declined

I think the second method is going to be too much hassle so I would go with the first.

dehall: Yeah I think passing the flag down to the ExpandedNode makes more sense.

ms-k1ngk0ng: Got it, thank you for the feedback! Will modify and re-push.

src/components/Node/Node.module.scss
```diff
@@ -28,6 +28,10 @@
     background-color: $red;
   }
 
+  &.declined {
+    background-color: #404040;
```
radamson: I'm not sure if this color is used elsewhere, but it would be better to keep this color with the rest in `src/styles/_variables.scss` sooner than later.

Maybe could call it `$black-medium`?

src/components/Node/Node.module.scss
```diff
@@ -60,4 +64,8 @@
     border-width: 1px;
     border-style: solid;
   }
+
+  &.childDeclined {
+    border: 1px solid #404040
```
blangley28: Use `$black-medium` here too
```suggestion
    border: 1px solid $black-medium
```

blangley28 (approved):

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -128,28 +130,32 @@ const ExpandedNodeField: FC<ExpandedNodeFieldProps> = ({ title, description }) =
...
   const rawDate = documentation.resource?.meta?.lastUpdated;
-  if (rawDate)
+  if (rawDate) {
+    const date = new Date(rawDate).toLocaleString('en-us');
     return (
       <ExpandedNodeField
         key="Status"
-        title={status}
-        description={new Date(rawDate).toLocaleString('en-us')}
+        title={isAccepted === false ? 'Declined' : status}
```
noranda: This can be simplified. We can also make sure to capitalize the status with the following:

```suggestion
        title={isAccepted ? status.charAt(0).toUpperCase() + status.slice(1) : 'Declined'}
```

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -128,28 +130,32 @@ const ExpandedNodeField: FC<ExpandedNodeFieldProps> = ({ title, description }) =
...
+        description={isAccepted === false ? date.concat(' by Dr. Example') : date}
```
noranda: Same here.

```suggestion
        description={isAccepted ? date : date.concat(' by Dr. Example')}
```

src/components/Node/Node.module.scss
```diff
@@ -10,7 +10,7 @@
   justify-content: center;
```
noranda, line 13: The styles in the node header don't quite match the design. To fix, add the following to `.node { ... }`:
```css
font-size: 1.2em;
white-space: nowrap;
```

src/components/Node/Node.module.scss
```diff
@@ -10,7 +10,7 @@
...
 
   .icon {
-    font-size: 1.2em;
+    font-size: 1.7em;
     color: $black;
     padding-right: 5px;
   }
```
noranda, line 21: As above, we need to fix a few styles to get the node header more in line with the design. First, replace this with:

```css
.icon {
  color: $black;
  margin-right: 8px;
}
```

Then add:

```css
.statusIcon {
  margin-left: 20px;
}
```

Then replace the `StatusIcon` function in the Node component with the following:

```js
const StatusIcon: FC<StatusIconProps> = ({ status }) => {
  if (status === null) return null;
  const icon = status ? faCheckCircle : faTimesCircle;
  return <FontAwesomeIcon icon={icon} className={styles.statusIcon} />;
};
```

These changes affect the comments section in the ExpandedNode component, so we also need make a few small changes to fix them. First, add a `<div>` around the label and button for the comments like so:

```js
<div>
  <label>Comments:</label>
  <button
    className={`${indexStyles.button} ${styles.defaultTextButton}`}
    onClick={(e): void => {
      e.preventDefault();
        if (!comments.includes(defaultText)) setComments(comments + defaultText);
    }}
  >
    Use Default Text
  </button>
</div>
```

And finally, fix the `.commentsForm` css selector in the stylesheet for the ExpandedNode component as follows:

```css
.commentsForm {
  margin-top: 15px;

  > div > label {
    font-weight: bold;
  }
}
```

src/styles/_nodes.scss
```diff
@@ -35,6 +35,8 @@
 .iconAndLabel {
   flex-grow: 1;
   text-align: center;
+  display: flex;
+  justify-content: center;
 }
```
noranda, line 41: Add `align-items: center;` to `.iconAndLabel { ... }` and replace the `.statusIcon` styles with:

```css
.statusIcon {
  margin: 0;
  padding-left: 20px;
}
```

noranda (changes requested):
This looks pretty good! Nice job!

There are a few styling changes that I think we should make to the node headers to get them more in line with the design. I've made a few comments with suggested changes to handle that. Let me know if you have any questions.

noranda (approved):

PR #50, opened 2020-05-26
Add Careplan

KeeyanGhoreshi:
A lot less change was needed than I thought.

### DIFF
--- a/public/static/pathways/her2_pathway.json
+++ b/public/static/pathways/her2_pathway.json
@@ -68,17 +68,8 @@
           "type": "create",
           "description": "Chemotherapy TH Weekly",
           "resource": {
-            "resourceType": "ServiceRequest",
-            "code": {
-              "coding": [
-                {
-                  "system": "http://snomed.info/sct",
-                  "code": "367336001",
-                  "display": "Chemotherapy (procedure)"
-                }
-              ],
-              "text": "Chemotherapy (procedure)"
-            }
+            "resourceType": "CarePlan",
+            "title": "ChemotherapyTH"
           }
         }
       ],
@@ -87,7 +78,7 @@
           "transition": "PostchemotherapyTrastuzumabRequest"
         }
       ],
-      "cql": "if exists [Procedure: \"Chemotherapy (procedure) code\"] then [Procedure: \"Chemotherapy (procedure) code\"] Chemo return Tuple{ resourceType: 'Procedure', id: Chemo.id.value , status: Chemo.status.value} else [ServiceRequest: \"Chemotherapy (procedure) code\"] Request return Tuple{ resourceType: 'ServiceRequest', id: Request.id.value, status: Request.status.value }"
+      "cql": "[CarePlan] CTH where CTH.title.value='ChemotherapyTH' return Tuple{ resourceType: 'CarePlan', id: CTH.id.value , status: CTH.status.value}"
     },
     "ChemotherapyTCHAC+TH": {
       "label": "Chemotherapy TCH AC + TH",
@@ -96,17 +87,8 @@
           "type": "create",
           "description": "Chemotherapy TCH AC + TH",
           "resource": {
-            "resourceType": "ServiceRequest",
-            "code": {
-              "coding": [
-                {
-                  "system": "http://snomed.info/sct",
-                  "code": "367336001",
-                  "display": "Chemotherapy (procedure)"
-                }
-              ],
-              "text": "Chemotherapy (procedure)"
-            }
+            "resourceType": "CarePlan",
+            "title": "ChemotherapyTCH"
           }
         }
       ],
@@ -115,7 +97,7 @@
           "transition": "PostchemotherapyTrastuzumabRequest"
         }
       ],
-      "cql": "if exists [Procedure: \"Chemotherapy (procedure) code\"] then [Procedure: \"Chemotherapy (procedure) code\"] Chemo return Tuple{ resourceType: 'Procedure', id: Chemo.id.value , status: Chemo.status.value} else [ServiceRequest: \"Chemotherapy (procedure) code\"] Request return Tuple{ resourceType: 'ServiceRequest', id: Request.id.value, status: Request.status.value }"
+      "cql": "[CarePlan] CTCH where CTCH.title.value='ChemotherapyTCH' return Tuple{ resourceType: 'CarePlan', id: CTCH.id.value , status: CTCH.status.value}"
     },
     "PostchemotherapyTrastuzumabRequest": {
       "label": "Post-Chemotherapy Trastuzumab Request",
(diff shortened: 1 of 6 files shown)


dehall (commented):
At a glance the engine changes look good. I'll go through that in more detail later.
Main thing though is that the CarePlan is intended for chemo regimens not all medications.

So, on the Early Stage HER2+ Pathway, "Chemotherapy TH Weekly" and "Chemotherapy TCH AC + TH" should both be CarePlans where today they are a ServiceRequest for a generic chemo procedure. Trastuzumab should stay a MedicationRequest because it's only one medication.  

On the Neoadjuvant chemo pathway, the 3 nodes with "Chemotherapy" in the name should all become CarePlans. On that pathway though there are 2 nodes with the names of 2 medications. Not sure if those are technically regimens - we should check with Jim on those.

Last thing is somewhere we need to change the icon for CarePlans. Not sure what a good icon to represent chemo regimens is. Maybe Noranda has some ideas.

public/static/pathways/her2_pathway.json
```diff
@@ -68,17 +68,8 @@
...
-                {
-                  "system": "http://snomed.info/sct",
-                  "code": "367336001",
-                  "display": "Chemotherapy (procedure)"
-                }
-              ],
-              "text": "Chemotherapy (procedure)"
-            }
+            "resourceType": "CarePlan",
+            "title": "ChemotherapyTH"
```
blangley28, line 72: So it looks like we are using the title for the CQL but should we update to actually include the correct information for the medication requests?

blangley28: From the sprint review meeting yesterday it sounds like this was the goal of the task so no need to update anything

blangley28 (commented):
So the engine is a complex nightmare but I believe the intended flow for approving is:
1) approval of guidance node creates the request
2) node is marked as accepted
3) once the actual event (procedure/medication etc) is complete in the patient record then we move on to the next node

So with careplan I think it becomes a bit more complex because we have the plan, request, and event. So I think when we accept a careplan it should write the careplan and any requests. Perhaps that is out of scope for this task.
I do think something fishy is going on here. I would have thought based on your changes the engine would progress to the next state but it is actually stopping at the correct place. I'm not sure if it will progress correctly when the careplan has been executed though - mostly because I am not exactly sure what needs to be in the patient record for that to happen....

public/static/pathways/her2_pathway.json
dehall: Yeah we have a task in the backlog to fill out more details on the CarePlan, but that's going to involve more UI work so I think that's going to be on the back burner for a while.

public/static/pathways/neoadjuvant_pathway.json
```diff
@@ -240,6 +248,14 @@
...
         "action": [
+          {
+            "type": "create",
+            "description": "Chemo Care Plan",
+            "resource": {
+                "resourceType": "CarePlan",
+                "title": "ChemoACTHP"
+            }
+          },
           {
```
dehall: Technically it doesn't hurt to leave these MedicationRequests in since they don't do anything, but on the other hand maybe we want to remove them so that people don't think they do something. Eventually we will want to migrate these into the CarePlan resource itself, as activities of that care plan.

public/static/pathways/neoadjuvant_pathway.json
```diff
@@ -240,6 +248,14 @@
...
             "type": "create",
```
dehall: Also GitHub won't let me add a comment on the actual line, but the CQL here should look for the completed CarePlan not the medication

KeeyanGhoreshi:
The CarePlan follows the same rule as the service/medication requests.  It has an overall `status` property which allows the pathway to move on when it is set to `completed`.

dehall (commented):
Yeah this looks good in terms of the engine changes. Ben your sequence above is correct, but the other thing to keep in mind is that Camino works in conjunction with the EHR, and shouldn't need to handle everything itself. In the case of a chemotherapy regimen, we want to consider the regimen an atomic unit to be written back to the EHR, which can handle the internal orders in the appropriate sequence.

public/static/pathways/neoadjuvant_pathway.json
blangley28: Yea we probably should remove them. If we do though can we either link to this version in the update careplan ticket or copy them over to that task (just so whoever picks it up can quickly see what the medications are instead of having to look up the regimens again)

blangley28:
> Ben your sequence above is correct, but the other thing to keep in mind is that Camino works in conjunction with the EHR, and shouldn't need to handle everything itself.

Are you saying Camino shouldn't need to make the requests or the events? I guess my bigger comment was more about checking if the CarePlan is complete. Is marking the status as complete sufficient

dehall:
In the case of the chemo regimen CarePlan, yeah I don't think we should worry about breaking out medication requests from subcomponents of the CarePlan. I mean maybe that's something to think about down the line in a production application, but for now we should just consider a CarePlan to be a request, write the one top-level request to the record on accept, and then look at the status field on the resource to determine if it's complete. That way we don't need any further engine changes

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -312,6 +318,13 @@ function renderGuidance(
         }
         break;
       }
+      case 'CarePlan': {
+        const plan = documentation.resource as CarePlan;
+        const title = plan.title;
+        if (title) {
+          returnElements.push(<ExpandedNodeField key="Title" title="Title" description={title} />);
```
radamson, line 332: should the title property actually be "Title" or should it be `title`?  Looks weird to me.

```suggestion
          returnElements.push(<ExpandedNodeField key="Title" title={title} description={title} />);
```

blangley28: Wouldn't that make the title and description the same thing?

KeeyanGhoreshi: the title is more like the key, and the key is just the key.

radamson: The explanation didn't help me, but looking at the `ExpandedNodeField` I see that title is really the field's title.  I'm going to assume that we'll never be pushing more than one `ExpandedNodeField` of this kind (i.e. with the same key).

https://github.com/mcode/pathways/blob/d4dd29451472d7bf840a2c3e1643fcfd30abbbea/src/components/ExpandedNode/ExpandedNode.tsx#L120-L127

blangley28 (approved):

public/static/pathways/neoadjuvant_pathway.json
```diff
@@ -242,91 +182,14 @@
...
-                  }
-                ],
-                "text": "trastuzumab"
-              }
-            }
-          },
-          {
-            "type": "create",
-            "description": "Paclitaxel Medication request",
+            "description": "Chemo Care Plan",
```
dehall: Can you update this one to say "Chemotherapy ACTHP Regimen"? Also since the Type field in the UI already says "care plan" also could you rename the other ones to say Regimen as well? That's less duplicative 
image

dehall (commented):
Looks good! Just one comment and it needs a rebase then I'll merge

public/static/pathways/neoadjuvant_pathway.json
KeeyanGhoreshi: There is already a Chemotherapy ATCHP, are there meant to be 2?

dehall: This state already has the label Chemotherapy ATCHP but the description is just generic. See the middle one stands out
image

dehall (approved):

PR #51, opened 2020-05-28
Updated nodes of type "ServiceRequest" with an alternate icon.

ms-k1ngk0ng:
Also reviewed other icon options, all of which seemed appropriate. The similarity between "MedicationRequest" and "MedicationAdministration" might be a tad confusing and I would suggest changing the "MedicationAdministration" icon.
Screen Shot 2020-05-28 at 9 43 44 AM

### DIFF
--- a/src/components/Node/Node.tsx
+++ b/src/components/Node/Node.tsx
@@ -13,11 +13,9 @@ import {
   faMicroscope,
   faPlay,
   faPrescriptionBottleAlt,
-  faCapsules,
   faSyringe,
   faCheckCircle,
-  faTimesCircle,
-  faClinicMedical
+  faTimesCircle
 } from '@fortawesome/free-solid-svg-icons';
 
 const useStyles = makeStyles(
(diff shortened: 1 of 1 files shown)


ms-k1ngk0ng:
Here are the other icon options I was considering - let me know if you think any of them would work better than the one I selected!

ms-k1ngk0ng:
Screen Shot 2020-05-28 at 9 45 45 AM

ms-k1ngk0ng:
Screen Shot 2020-05-28 at 9 48 47 AM

ms-k1ngk0ng:
Screen Shot 2020-05-28 at 9 49 39 AM

blangley28 (commented):
Looks good to me from a code perspective. I think some of the test pathways still have Procedure but those should all be ServiceRequests. Perhaps we should use the same icon in both cases and then make a backlog item to go through and update all the instances of Procedure

dehall:
Yeah as Ben mentioned, there shouldn't still be any instances of Procedure in the pathways, and actually there shouldn't be any MedicationAdministration either. (The idea is that Camino recommends the action and orders it in the EHR system if accepted, but the details of how that care is provided is left to the normal EHR.)
So, let's remove the Procedure and MedicationAdministration icons here and make a separate task to clean those up in the pathways.

I don't have strong opinions on which icons we should actually use here, but if it helps, the ServiceRequests represent a request for a procedure to be performed, which in the case of cancer care is likely to mean a surgery or radiotherapy or something along those lines. So I don't know if a syringe or the doctor with stethoscope or something else is better for that.

blangley28:
> So, let's remove the Procedure and MedicationAdministration icons here and make a separate task to clean those up in the pathways.

Created https://jira.mitre.org/browse/PATHWAYS-281 to address the clean up

ms-k1ngk0ng:
Okay, I removed assignment of icons for "Procedure" and "MedicationAdministration", so any nodes that have those types by default receive the microscope icon. I also decided to switch the ServiceRequest icon to the syringe, now that it is no longer being used for "Procedure".

ms-k1ngk0ng:
Screen Shot 2020-05-28 at 4 10 37 PM

blangley28 (approved):

noranda:
This looks good to me. I like sticking with the syringe icon for procedures for now, but we can follow up on this when we start getting some user feedback.

noranda (approved):

PR #52, opened 2020-05-29
Make notes visible for actioned nodes.

ms-k1ngk0ng:


ms-k1ngk0ng:
Screen Shot 2020-05-29 at 3 26 32 PM

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -348,6 +348,9 @@ const ExpandedNodeMemo: FC<ExpandedNodeMemoProps> = memo(
           <tbody>
             <StatusField documentation={documentation} />
             {guidance || branch}
+            {!isActionable && comments !== '' && comments != null && (
```
blangley28: Since `''` and `null` are falsy you can condense the last two.
```suggestion
            {!isActionable && comments && (
```

blangley28 (commented):
So I think we are going to need to grab the comments from the DocumentReference resource written back to the patient record. The `comments` variable here is part of the `ExpandedNode` react state. That means whenever the component is unmounted (such as refreshing the page or coming back at a later date) the comments will be lost. To see what I mean select Early Stage HER2+ Pathway and add comments to the Chemotherapy. Either return to Explore Pathways and back to this pathway or toggle between the two pathways. Once the Early Stage HER2+ Pathway renders again you will notice the comments are gone.
The Jira ticket should have some hints on where to search for code in the engine which pulls out the DocumentReference resource

noranda (commented):
Agree with @blangley28's comment on making sure the comments persist. Looks good from a styling perspective.

ms-k1ngk0ng:
I updated this PR with known problems, primarily having to do with the DefaultText button. 1) Clicking the Use Default Text button does save the Default Text as a comment, however it does not fill the textarea box upon clicking. 2) Clicking Use Default Text overwrites whatever information is written in the textarea box. I think clicking Use Default Text should append to whatever is written in the box, not overwrite it. 3) A blank comment still gets stored - not a major problem and I think a quick fix once the other stuff gets sorted out. I'm not sure what exactly is causing these issues, but I would guess it has something to do with calling setComments not altering the notes value I pass down. I'm not sure what the best way is to address these issues, so open to any feedback!

src/engine/output-results.ts
```diff
@@ -12,7 +12,8 @@ import {
   CriteriaResultItem,
   Documentation
 } from 'pathways-model';
-import { DocumentReference, DomainResource } from 'fhir-objects';
+import { DomainResource } from 'fhir-objects';
```
blangley28: I don't think DocumentReference should have been removed here. I'm getting an error when I run this because it is used on line 235
```suggestion
import { DocumentReference, DomainResource } from 'fhir-objects';
```

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -98,6 +99,16 @@ const ExpandedNode: FC<ExpandedNodeProps> = memo(
...
 
+    let notes;
+    const documentreference = retrieveNote(pathwayState.label, patientRecords);
+    if (documentreference) {
+      const content = documentreference.content[0].attachment?.data;
+      if (content) {
+        notes = atob(content);
+        notes = notes.slice(notes.indexOf('Notes: ') + 6);
+      }
+    }
```
blangley28: So I think your problem with the notes not being added to the textbox has to do with this block. On line 121 you pass in `notes` as the comments which is being used to fill in the textbox. We should probably still use the `note.notes` state to populate that comment section but then move this block so the `notes` you defined here are used in place of `comments` in line 409 below

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -425,7 +439,7 @@ const ExpandedNodeMemo: FC<ExpandedNodeMemoProps> = memo(
             </div>
             <textarea
               className={styles.comments}
-              value={comments}
+              defaultValue={comments}
```
blangley28: I think this should also remain value. Getting some unexpected behavior. Making this value and changing the comments prop back to `notes.note` means we will still see the comments in the box and the default text button will append correctly to the textbox
```suggestion
              value={comments}
```

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -107,7 +118,7 @@ const ExpandedNode: FC<ExpandedNodeProps> = memo(
           pathwayState={pathwayState}
           documentation={documentation}
           setComments={setComments}
-          comments={note?.notes ?? ''}
+          comments={notes ?? ''}
```
blangley28: I think changing this to 
```suggestion
          comments={notes ?? note?.notes ?? ''}
```
gives the desired behavior

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -122,6 +124,7 @@ const ExpandedNode: FC<ExpandedNodeProps> = memo(
           }}
           isAccepted={isAccepted}
           onAdvance={onAdvance}
+          patientRecords={patientRecords}
```
blangley28: Instead of passing this down can we use the PatientRecord context. Remove this line and 374 and then add the following at the beginning of the `ExpandedNodeMemo`
```
const { patientRecords } = usePatientRecords();
```

blangley28 (commented):
I don't think this is an issue with this PR but we may want to address it later - it looks like return characters are being ignored. If I write a multiline note the preview in the modal and the saved note in the expanded node have the return character removed. Not sure if it is because the character is being stripped out or not displayed

dehall:
Re: the return characters - that's a display issue. Return characters and whitespace are ignored in HTML so we would have to manually replace \n with `<br/>` or something along those lines

dehall:
Functionality looks great. A couple things:
1. The "extra characters" you mentioned at the end of the note string are inserted when the note is created -- see NoteDataProvider.tsx, the last 2 lines here:
```jsx
  const noteString = `Date: ${note.date}\n
Patient: ${note.patient}\n
Birthdate: ${note.birthdate}\n
Physician: ${note.physician}\n
${mcodeElements}
Pathway: ${note.pathway}\n
Node: ${note.node}\n
Status: ${note.status}\n
Treatment: ${note.treatment}\n
Notes: ${note.notes}
  `;
```
If we put the closing quote on the same line as Notes instead of on the next line it should clean those extra characters up.

2. The combination of "Notes" and "Comments" showing different things is a little confusing. The label that currently says "Notes" should really be "Description" since that's the underlying field in the data model that line is getting pulled from. Can you change that on lines 293/294 of ExpandedNode.tsx?
```jsx
    <ExpandedNodeField
      key="Notes" <--
      title="Notes" <--
      description={pathwayState.action[0].description}
    />,
```

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -383,17 +384,32 @@ const ExpandedNodeMemo: FC<ExpandedNodeMemoProps> = memo(
...
+      }
+    }
+
     return (
       <div className={indexStyles.expandedNode}>
         <table className={styles.infoTable}>
           <tbody>
             <StatusField documentation={documentation} isAccepted={isAccepted} />
             {guidance || branch}
+            {!isActionable && notes && /\S/.test(notes) && (
```
blangley28, line 410: Do we still need what I presume is the whitespace check here?
```suggestion
            {!isActionable && notes && (
```

ms-k1ngk0ng: I think so, in case someone tries to submit a comment that has nothing but a whitespace (i.e. accidentally pressed the spacebar or return) then this catches that and stops the comments title from being visible with a blank description box.

blangley28: Good enough to convince me

ms-k1ngk0ng: Screen Shot 2020-06-09 at 2 52 35 PM

blangley28 (approved):

dehall (approved):

PR #53, opened 2020-06-02
Assign/Unassign Pathway

blangley28:
Implement functionality to unselect a pathway. Update the assign/unassign UI as well as allow the user to unselect (unassign) a pathway from a patient.
Screen Shot 2020-06-02 at 4 13 35 PM

Screen Shot 2020-06-02 at 4 13 15 PM

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -204,13 +204,13 @@ const App: FC<AppProps> = ({ demoId }) => {
 
   const unassignPathway = useCallback(
     (pathwayName: string): void => {
-      const newPatientRecords = patientRecords.filter(r => {
-        return r.resourceType === 'CarePlan' && (r as CarePlan).title === pathwayName
-          ? false
-          : true;
+      const newPatientRecords: DomainResource[] = [];
+      patientRecords.forEach(r => {
+        if (r.resourceType === 'CarePlan' && (r as CarePlan).title === pathwayName)
+          client?.delete?.(`CarePlan/${r.id}`);
+        else newPatientRecords.push(r);
       });
       setPatientRecords(newPatientRecords);
-      client?.delete?.(`CarePlan/${pathwayName}`);
     },
     [patientRecords, setPatientRecords, client]
   );
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -17,8 +17,8 @@ import { EvaluatedPathway } from 'pathways-model';
 import useGetPathwaysService from './PathwaysService/PathwaysService';
 import FHIR from 'fhirclient';
 import { MockedFHIRClient } from 'utils/MockedFHIRClient';
-import { getHumanName } from 'utils/fhirUtils';
-import { DomainResource, Practitioner } from 'fhir-objects';
+import { getHumanName, createCarePlan } from 'utils/fhirUtils';
+import { DomainResource, Practitioner, CarePlan } from 'fhir-objects';
 import styles from './App.module.scss';
 import { UserProvider } from './UserProvider';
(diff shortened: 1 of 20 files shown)


src/components/App.tsx
```diff
@@ -190,6 +190,31 @@ const App: FC<AppProps> = ({ demoId }) => {
...
+    },
+    [patientRecords, patient, client, setPatientRecords]
+  );
+
+  const unassignPathway = useCallback(
+    (pathwayName: string): void => {
+      const newPatientRecords: DomainResource[] = [];
+      patientRecords.forEach(r => {
+        if (r.resourceType === 'CarePlan' && (r as CarePlan).title === pathwayName)
+          client?.delete?.(`CarePlan/${r.id}`);
```
dehall, line 210: For discussion only, no action requested: are there potential benefits/drawbacks to deleting the CarePlan vs marking it as `status: "revoked"` or possibly something else?

blangley28: I guess that depends if we are going to be populating this CarePlan with any details. If we are then it would make sense to keep it and just change the status

danlee1025:
image

The default text button looks like it overlaps the text box a little bit... Should we add some margin?

danlee1025:
image

When I hover over the unassign button in the pathway view it is still blue... Did we want it darker red like it looks on the explore pathway view?

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -240,36 +215,34 @@ const PathwaysListElement: FC<PathwaysListElementProps> = ({
...
-              disabled={selected}
+            <Button
               onClick={(): void => {
-                const carePlan = createCarePlan(pathway.name, patient);
-
-                setPatientRecords([...patientRecords, carePlan]);
-                client?.create?.(carePlan);
-                callback(evaluatedPathway);
+                if (selected) unassignPathway(pathway.name);
+                else assignPathway(pathway.name);
```
danlee1025: Might look cleaner here with a ternary instead of if else
```
selected ? unassignPathway(pathway.name) : assignPathway(pathway.name);
```

src/components/DropDown/DropDown.tsx
```diff
@@ -23,25 +29,47 @@ const DropDown: FC<Props> = ({
...
+  const { patientRecords } = usePatientRecords();
+  const { evaluatedPathway, assignPathway, unassignPathway } = usePathwayContext();
   const onChangeCallback = useCallback(
     (value: Option | ReadonlyArray<Option> | null | undefined) => {
       if (onChange) onChange(value == null ? null : value);
     },
     [onChange]
   );
 
+  const selected = pathwayIsSelected(patientRecords, evaluatedPathway?.pathway);
```
danlee1025: Do you think we should change this variable to `assigned`? I think this is partially my fault since when I first implemented "Select Pathway" I named everything `selected` instead of `assigned` but now that I see `DropDown.tsx` it can get quite confusing since it also uses the term select. If you think it's a good idea and it's not too much work, we should change the helper functions like `getSelectedPathways` and `pathwayIsSelected` as well as some of the variables in `PathwaysList`.

blangley28: Renamed everywhere I could find including the helper functions

blangley28:
> image
> 
> The default text button looks like it overlaps the text box a little bit... Should we add some margin?

Screen Shot 2020-06-03 at 1 58 44 PM

Added a margin to the comments box to provide a little gap

dehall (approved):

danlee1025 (approved):

PR #54, opened 2020-06-03
Force move to next state

danlee1025:
Added advance button to force move to next state when a node has been accepted/declined.

image

The advance button appears when the node is current, non-actionable, guidance, and contains transitions.

When button is clicked, a `DocumentReference` is added to the list of patient resources. The content in the `DocumentReference` just has a string that contains the state label and the word Advance, which is how I am searching for the `DocumentReference` in the engine. Did we want anything else in the content or the resource?

### DIFF
--- a/src/components/ExpandedNode/ExpandedNode.tsx
+++ b/src/components/ExpandedNode/ExpandedNode.tsx
@@ -90,13 +90,11 @@ const ExpandedNode: FC<ExpandedNodeProps> = memo(
     };
 
     const onAdvance = (): void => {
-      const newPatientRecords = [...patientRecords];
       const content = `${pathwayState.label} - Advance`;
       const documentReference = createDocumentReference(content, patient);
 
-      newPatientRecords.push(documentReference);
       client?.create?.(documentReference);
-      setPatientRecords(newPatientRecords);
+      setPatientRecords([...patientRecords, documentReference]);
     };
 
     return (
--- a/src/components/ExpandedNode/ExpandedNode.tsx
+++ b/src/components/ExpandedNode/ExpandedNode.tsx
@@ -12,6 +12,7 @@ import { usePatientRecords } from 'components/PatientRecordsProvider';
 import { usePatient } from 'components/PatientProvider';
 import {
   translatePathwayRecommendation,
+  createActionDocumentReference,
   createDocumentReference,
   createNoteContent
 } from 'utils/fhirUtils';
@@ -63,7 +64,12 @@ const ExpandedNode: FC<ExpandedNodeProps> = memo(
           note?.notes ?? '',
           pathwayState
         );
-        const documentReference = createDocumentReference(noteString, pathwayState.label, patient);
+        const documentReference = createActionDocumentReference(
+          noteString,
+          pathwayState.label,
+          patient
+        );
+
         newPatientRecords.push(documentReference);
         client?.create?.(documentReference);
       }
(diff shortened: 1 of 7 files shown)


src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -81,11 +88,22 @@ const ExpandedNode: FC<ExpandedNodeProps> = memo(
...
     };
 
+    const onAdvance = (): void => {
+      const newPatientRecords = [...patientRecords];
+      const content = `${pathwayState.label} - Advance`;
+      const documentReference = createDocumentReference(content, patient);
+
+      newPatientRecords.push(documentReference);
+      client?.create?.(documentReference);
+      setPatientRecords(newPatientRecords);
```
blangley28: In other places I think we use this to be more concise
```suggestion
      setPatientRecords([...patientRecords, documentReference]);
```

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -373,12 +396,18 @@ const ExpandedNodeMemo: FC<ExpandedNodeMemoProps> = memo(
             {guidance || branch}
           </tbody>
         </table>
+        {/* Node is advanceable if it has been accepted or declined */}
+        {pathwayState.transitions.length > 0 && !isActionable && isGuidance && isCurrentNode && (
+          <button className={`${indexStyles.button} ${styles.button}`} onClick={onAdvance}>
+            Advance
+          </button>
```
blangley28: We are switching to the material-ui button to be consistent with the builder
```suggestion
          <Button 
               className={`${indexStyles.button} ${styles.button}`} 
               onClick={onAdvance}
               variant="contained"
               color="primary"
          >
            Advance
          </Button>
```

src/components/Node/Node.tsx
```diff
@@ -121,6 +121,7 @@ const Node: FC<NodeProps & { ref: Ref<HTMLDivElement> }> = memo(
                 isGuidance={isGuidance}
                 documentation={documentation}
                 isAccepted={isAccepted}
+                isCurrentNode={isCurrentNode}
```
blangley28, line 124: We're passing this around alot. It seems like it is computed in the Graph component. The evaluatedPathway is available through the pathway context so I think we should be able to compute this in the ExpandedNodeMemo instead of passing it all the way through.

src/engine/output-results.ts
```diff
@@ -221,14 +221,40 @@ function isComplete(resource: DocumentationResource): boolean {
...
 }
 
+/**
+ * Helper function to determine whether current state should be skipped and pathway execution should advance
+ * Checks for DocumentReference resource with state label and advance
+ * @param currentState current state
+ * @param resources list of patient resources
+ * @return boolean as to whether pathway execution should advance
+ */
+function shouldAdvance(currentState: State, resources: DomainResource[]): boolean {
```
blangley28, line 231: Nice! I appreciate the docs too!

blangley28 (changes requested):
Few minor comments but it looks good

src/engine/output-results.ts
```diff
@@ -221,14 +221,40 @@ function isComplete(resource: DocumentationResource): boolean {
...
+  return resources.some(r => {
+    if (r.resourceType !== 'DocumentReference') return false;
+    const content = (r as DocumentReference).content[0].attachment.data;
+    if (content) {
+      const convertedContent = atob(content);
+      return convertedContent.includes(currentState.label) && convertedContent.includes('Advance');
```
dehall: If there's a specific format that this text is always going to be, can we have this check for that exact format? Only concern is that the label of the state could contain the word "Advance" or "Advanced" and conflict.

danlee1025: Oh I had not thought about that concern, I'll update to check for the exact string.

src/components/ExpandedNode/ExpandedNode.tsx
danlee1025: Thanks, I think I was copying how the default text button was styled... I'll update that button to material-ui also.

blangley28: Don't worry about updating the other buttons - I updated them in my PR

src/components/Node/Node.tsx
danlee1025: Hmm so I looked into this. We do have access to the pathway from context but we don't have access to the `nodeName`(which seems like we need to calculate `isCurrentNode`). We do have access to the state label and it seems like `nodeName` is just a trimmed version of the label but I would rather not make that assumption and just pass the the `isCurrentNode` prop. What do you think?

blangley28: Yea you're right. The `nodeName` is the key which can actually be anything. Lets just leave it as you have it.

blangley28 (approved):

dehall:
I'm not too worried about it, but the button now has rounded corners whereas every other button has square corners. If that's quick to change great, otherwise I'll approve and merge this later this afternoon

dehall:
Oh wait, the assign/unassign pathway PR changed the other buttons to be rounded too.

dehall (approved):

PR #56, opened 2020-06-04
Fix CQL issues on MedicationRequests

dehall:
Fixes a couple CQL issues that were preventing MedicationRequest nodes from accepting correctly.

1) Regenerated ELM on the Early Stage HER2+ pathway
2) Added missing RxNorm codes to the mCODE Library CQL, and updated the relevant nodes on the Neoadjuvant pathway to point to those, and regenerated the ELM.

With this every node on both pathways should correctly show as accepted or declined based on the user's action


ms-k1ngk0ng (approved):
Not a 100% sure what everything is doing so you may want to request a third reviewer but from my end it looks good!

danlee1025 (approved):

PR #58, opened 2020-06-09
Nodes red

ms-k1ngk0ng:


ms-k1ngk0ng:
Screen Shot 2020-06-09 at 1 38 56 PM

Realized that I had been editing my other PR (actionable-notes) on this branch, which is why there are a ton of commits having to do with retrieve note. I just went through and deleted everything having to do with actionable-notes from this branch, after double-checking that I have a local branch of that PR. Everything on the most recent push has to do only with nodes-red.

PR #59, opened 2020-06-19
alter behavior when node is guidance node

KeeyanGhoreshi:
I'm not sure if this is the intended way we want to fix this bug, since it doesn't really change the fact that it's still treating branch nodes having missing data filled as being "declined", but it does alter the behavior branch nodes to show "Status" instead of "declined".  I think the main issue was that "isAccepted" is null for branch nodes, so when it went to fill out the "status" field it just defaulted to using "declined" instead, since null would be treated as false.

### DIFF
--- a/src/components/ExpandedNode/ExpandedNode.tsx
+++ b/src/components/ExpandedNode/ExpandedNode.tsx
@@ -162,18 +162,18 @@ const StatusField: FC<StatusFieldProps> = ({ documentation, isAccepted }) => {
   const rawDate = documentation.resource?.meta?.lastUpdated;
   if (rawDate) {
     const date = new Date(rawDate).toLocaleString('en-us');
-    let declinedText = '';
+    let titleText = '';
     if (isAccepted) {
-      declinedText = status.charAt(0).toUpperCase() + status.slice(1);
+      titleText = status.charAt(0).toUpperCase() + status.slice(1);
     } else if (isAccepted === null) {
-      declinedText = 'Status';
+      titleText = 'Status';
     } else {
-      declinedText = 'Declined';
+      titleText = 'Declined';
     }
     return (
       <ExpandedNodeField
         key="Status"
-        title={isAccepted ? status.charAt(0).toUpperCase() + status.slice(1) : declinedText}
+        title={titleText}
         description={isAccepted ? date : date.concat(' by Dr. Example')}
       />
     );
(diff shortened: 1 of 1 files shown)


src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -152,20 +152,22 @@ const ExpandedNodeField: FC<ExpandedNodeFieldProps> = ({ title, description }) =
...
   const status = documentation.status;
   const rawDate = documentation.resource?.meta?.lastUpdated;
   if (rawDate) {
     const date = new Date(rawDate).toLocaleString('en-us');
+    const declinedText = isGuidance ? 'Declined' : 'Status';
     return (
       <ExpandedNodeField
         key="Status"
-        title={isAccepted ? status.charAt(0).toUpperCase() + status.slice(1) : 'Declined'}
+        title={isAccepted ? status.charAt(0).toUpperCase() + status.slice(1) : declinedText}
```
dehall: Is there even possibly a simpler approach? `isAccepted` is `boolean | null`, so I think it's that `null` case that is tripping up the logic here. Extra parameters might not even be needed

| `isAccepted` | expected result |
|--------------|---------|
| `true` | "Accepted" <br/>(or I guess the `status` field from the resource like we already have) |
| `false` | "Declined" |
| `null` | "Status" |

dehall (approved):

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -162,10 +162,18 @@ const StatusField: FC<StatusFieldProps> = ({ documentation, isAccepted }) => {
...
+    } else if (isAccepted === null) {
+      declinedText = 'Status';
+    } else {
+      declinedText = 'Declined';
+    }
     return (
       <ExpandedNodeField
         key="Status"
-        title={isAccepted ? status.charAt(0).toUpperCase() + status.slice(1) : 'Declined'}
+        title={isAccepted ? status.charAt(0).toUpperCase() + status.slice(1) : declinedText}
```
danlee1025: This line can probably just be `title={declinedText}` since you are already checking the `isAccepted` variable when you instantiate the value.  This can also probably be renamed to something like `titleText` since the text is not only for declined nodes.

danlee1025 (approved):

PR #61, opened 2020-07-09
Added checkmark to assigned pathways.

ms-k1ngk0ng:
Visible on the dropdown menu within the pathway view.

### DIFF
--- a/src/components/DropDown/DropDown.tsx
+++ b/src/components/DropDown/DropDown.tsx
@@ -40,6 +40,17 @@ const DropDown: FC<Props> = ({
 
   const assigned = pathwayIsAssigned(patientRecords, evaluatedPathway?.pathway);
 
+  const formatOptionLabel = (option: Readonly<Option>): object => (
+    <div>
+      <div className={styles.name}> {option.label} </div>
+      <div className={styles.check}>
+        {pathwayIsAssigned(patientRecords, option.value.pathway) && (
+          <FontAwesomeIcon icon={faCheckCircle} />
+        )}
+      </div>
+    </div>
+  );
+
   if (visible)
     return (
       <div className={styles.dropdown}>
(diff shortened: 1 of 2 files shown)


ms-k1ngk0ng:
Neither pathway assigned.
Screen Shot 2020-07-09 at 10 31 05 AM
One pathway assigned.
Screen Shot 2020-07-09 at 10 31 10 AM
Both assigned.
Screen Shot 2020-07-09 at 10 31 18 AM

src/components/DropDown/DropDown.tsx
```diff
@@ -40,6 +40,17 @@ const DropDown: FC<Props> = ({
 
   const assigned = pathwayIsAssigned(patientRecords, evaluatedPathway?.pathway);
 
+  const formatOptionLabel = (option: Readonly<Option>): object => (
+    <div>
```
noranda: You can use React fragments here (`<> ... </>`) instead of an empty div which doesn't serve a purpose here.

src/components/DropDown/DropDown.tsx
```diff
@@ -40,6 +40,17 @@ const DropDown: FC<Props> = ({
...
+      <div className={styles.name}> {option.label} </div>
+      <div className={styles.check}>
```
noranda: This div is also unnecessary. I would remove it and add the class directly to the `FontAwesomeIcon` component. Then you can delete the styles you have for `check` and just add `margin-left: 5px;`. You can also get rid of the `<div>` tag for the label, which would make the `name` class name with `display: inline-block;` unnecessary since the label is treated as an inline-block by default.

```
const formatOptionLabel = (option: Readonly<Option>): object => (
  <>
    {option.label}
    {pathwayIsAssigned(patientRecords, option.value.pathway) && (
      <FontAwesomeIcon icon={faCheckCircle} className={styles.check} />
    )}
  </>
);
```

```
.check {
  margin-left: 5px;
}
```

noranda (changes requested):
Looks good! The functionality is working as expected. I only have a few minor requested changes in the code.

noranda (approved):

dehall (approved):

PR #66, opened 2020-09-11
Temporary fix for Chrome and FF bug with flex column wrapping

noranda:
This PR is a temporary fix for the Chrome and Firefox bug we are seeing with flex column wrapping. This is a known [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=507397) that hasn't been fixed, but this is the workaround. This is needed for an upcoming demo.

Note: This fixes the issue on Chrome and Firefox, but seems to mess with the line-height in Firefox.

I recommend we create a new task to come up with a better fix for this, perhaps using tables instead of flexbox here.

### DIFF
--- a/src/components/ReportModal/ReportSection/ReportSection.module.scss
+++ b/src/components/ReportModal/ReportSection/ReportSection.module.scss
@@ -2,9 +2,9 @@
 
 .fieldContainer {
   display: flex;
-  flex-direction: column;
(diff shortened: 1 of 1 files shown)


dehall (approved):
Confirmed it looks good in Chrome, nothing else impacted. FF is a little weird as noted but that's ok for now

KeeyanGhoreshi (approved):

PR #67, opened 2020-10-21
Update Camino to use pathway-builder data model

dehall:
Updates Camino to use the data model from the pathway-builder. This includes not just the tweaks that have been made to the data model over the past few months, but more importantly the conceptual change in branch criteria. Previously the CQL for branch criteria would return a FHIR resource as documentation (or more accurately, a tuple with the resourceType and ID so that we could fetch the real resource when needed); now the CQL will just return a boolean. The impact here is that certain UI fields can no longer contain any details about the underlying resource because we just don't have a way of knowing what that resource is.

Some notes on  this PR:
 - The biggest change to the "evaluation engine" is in `output-results.ts` which turns the ELM results and list of patient resources into a current location in the pathway, plus path to get there. Previously this was very dependent on "documentation" which in most cases we may not have anymore. Rather than try to change this I'm trying to rewrite the main logic in a simpler way.
 - The version of the cql-execution library we were using does not support FHIR 4.0.1, only 4.0.0. Note that cqf-ruler only supports 4.0.1 so that should be the target. I updated that library here, which was originally supposed to be part of PATHWAYS-380.
 - There are too many TODOs in the code to even remember. Some of these will need to be addressed for this PR, some we can kick down the road to later.
 - We don't have pathway preconditions in exported pathways, but I updated the static examples to make sure we still had those. They now work the similarly to criteria, in that they are defined in one of the libraries, and bubble up to the main CQL library so they can be evaluated
 - I removed the "index.test.ts" file which is an integration test for the engine as a whole, because it was too difficult to reason about whether individual tests are correct or not. It imports a large pathway, a large patient record, runs the evaluation process, and compares against a large pre-computed result. Is that result actually correct? Or did we just run the process once, say "looks good", and say that's our expected result.


blangley28 (commented):
Still looking through the output-results changes but one thing I noticed is that action nodes are always marked as declined even if you accept

dehall:
> Still looking through the output-results changes but one thing I noticed is that action nodes are always marked as declined even if you accept

Yeah that's an artifact of broken CQL for the action nodes. If the CQL doesn't return a resource, then the display code interprets that as "completed + no resource == declined". I think it should work correctly with my sample testing pathways; in case I didn't update already here's the latest for those: 
[Archive.zip](https://github.com/mcode/pathways/files/5496495/Archive.zip)

blangley28:
Still need to look through the code thoroughly but using the pathways updated on #68 everything works correctly as far as I can tell

src/engine/output-results.ts
```diff
@@ -25,64 +24,199 @@ interface NodeData {
...
+  let currentNodes: string[] = [];
+  let nextNodes: string[] = [startNode];
+  const documentation: { [key: string]: Documentation } = {};
+
+  // think of this function like a proof by induction:
+  // - basis: we start at the start node
+  // - recursive: we know the current node(s) we are on, find the next ones
+  // if there are next ones, repeat the process with those, until we don't find any next steps
+  // then our "current node(s)" at the end, are the current step shown to the user
+
```
blangley28, line 44: I put a debugger right here and I noticed the `patientData` object (returned from executing the elm` is sometimes an empty list instead of `false`

dehall: The whole `patientData` object is a list? Or the fields within the object are sometimes a list? Originally the CQL in the pathway would always produce tuples/pseudo-resources, which if none was returned then we would get an empty list. The shift to boolean-valued criteria actually only changes the branch transition criteria, the CQL for action nodes should still return that pseudo-resource, so if an action is not present on the record it will return an empty list in the patient results. (If you break at the debugger again after accepting an action, it should come back with something in the list)

blangley28: Ahhh that makes sense

blangley28 (approved):
Everything looks good to me. Put the app through a stress test via the smart launcher and the demo endpoint. Nothing broke!

KeeyanGhoreshi:
I found a couple of UI bugs while testing but they're all present in the master branch as well, nothing breaking just some small inconsistencies.  The pathway for HER2 also seems to be missing a transition condition, which causes one of the options for `missing data` to be a blank button.  That also doesn't seem to originate from this pull request.  

Otherwise everything looked good.  Don't think any of those issues should hold up this PR.

KeeyanGhoreshi (approved):

